Issue #18, opened 2015-06-28
Nikon makernote "exposure difference" value


drewnoakes:
Diff between Java and C# versions:

```diff
diff --git a/jpg/metadata/nikon d610.jpg.txt b/jpg/metadata/nikon d610.jpg.txt
index 73e1db8..e944174 100644
--- a/jpg/metadata/nikon d610.jpg.txt	
+++ b/jpg/metadata/nikon d610.jpg.txt	
@@ -69,7 +69,7 @@ FILE: Nikon D610.jpg
 [Nikon Makernote - 0x000b] White Balance Fine = 0 0
 [Nikon Makernote - 0x000c] White Balance RB Coefficients = 458/256 330/256 256/256 256/256
 [Nikon Makernote - 0x000d] Program Shift = 0 EV
-[Nikon Makernote - 0x000e] Exposure Difference = -0.5 EV
+[Nikon Makernote - 0x000e] Exposure Difference = 20.83 EV
 [Nikon Makernote - 0x0011] Preview IFD = 13446
```

These values are quite different. Find out why and fix the broken implementation(s).

```diff
diff --git a/jpg/metadata/nikon d70.jpg.txt b/jpg/metadata/nikon d70.jpg.txt
index 8dcf2d9..ed40f03 100644
--- a/jpg/metadata/nikon d70.jpg.txt	
+++ b/jpg/metadata/nikon d70.jpg.txt	
@@ -67,7 +67,7 @@ FILE: Nikon D70.jpg
 [Nikon Makernote - 0x000b] White Balance Fine = 0
 [Nikon Makernote - 0x000c] White Balance RB Coefficients = 10242/1 34305/1 0/1 0/1
 [Nikon Makernote - 0x000d] Program Shift = 0 EV
-[Nikon Makernote - 0x000e] Exposure Difference = -1.5 EV
+[Nikon Makernote - 0x000e] Exposure Difference = 19.83 EV
 [Nikon Makernote - 0x0011] Preview IFD = 1379
 [Nikon Makernote - 0x0012] Auto Flash Compensation = 0 EV
 [Nikon Makernote - 0x0013] ISO = 0 400
```

drewnoakes:
Closed by @kwhopper in PR #26.

PR #26, opened 2015-11-26
cast Nikon EV byte array values to sbyte (signed byte) to correct calc

kwhopper:
This commit corrects Issue #18. Other GetEvDescription calls were tested, but should be verified. The double cast on values[2] may not be necessary but left it just in case.


drewnoakes:
Great fix. Thanks very much!

Issue #24, opened 2015-11-07
PCL Support


ravensorb:
Any thought to adding support for other platforms (like iOS and Android or even better PCL)?

drewnoakes:
That sounds great. Do you know what's required for these platforms? I have
no relevant experience myself.

ravensorb:
I did a bit of research and for the most part converting the library over should be easy with 1 exception -- XMP library that is being referenced is not PCL compatible.  I took a look at that library and it actually would not be too hard to convert as all that is needed is to switch from using System.Xml to System.Linq (XmlNode to XNode) and than it is fully PCL compatible.

drewnoakes:
The XMP library has been a source of many challenges for years in the Java version. I'm inclined to make it an optional dependency somehow, maybe through a plugin architecture.

As for iOS/Android, is there anything required beyond being a PCL library?

ravensorb:
I like the plugin in architecture -- maybe using DI/IoC?  And no, as long as the library is "porrtable" it works with Xamarin and as a result works with iOS/Android (As well as any other platform that supports .NET core -- like Windows IoT).

drewnoakes:
The simplest thing would be to create a PCL build that initially excludes
XMP support using #if PCL ... #endif

Later, if the XMP library becomes PCL compatible, it could be reintroduced.

drewnoakes:
Looking at this a little today I found a few challenges to porting. Nothing insurmountable, but they increase the time required to support PCL.

The following types are not available, but are used:

* `Encoding.ASCII` (use `Encoding.UTF8` instead?)
* `FileStream` (remove these APIs and only accept `Stream`?)
* `[Serializable]` and `SerializationInfo` (remove this)
* `IReadOnlyList<T>` (fall back to `IList<T>` as done in the NET35 build)
* `Console` (`Debug.WriteLine` instead?)
* `System.IO.Compression.DeflateStream` (use [this](https://www.nuget.org/packages/Microsoft.Bcl.Compression)?)

Also `Encoding.GetString(byte[])` doesn't exist, but is trivially satisfied by adding `offset` and `length` parameters.

drewnoakes:
Also `FileMetadataReader` has to go, as PCL has no abstraction of file systems, as far as I can tell.

drewnoakes:
Some work on the `pcl` branch. Pending are `Console`, `DeflateStream` and `Encoding.ASCII`. And of course, testing.

[...]

PR #27, opened 2015-12-19
Reinstate portable Xmp support (Issue #24) and add project to test Portable

kwhopper:
A local copy of XmpCore.Portable.dll is used. It should be replaced when/if the portable XmpCore gets a new nuget package. Tests can be run against each project independently, but two unsupported tests were removed from Portable.


kwhopper:
Two of the date functions fail unit tests, but they failed before these commits. The build seems fine now though. Sorry for extra commits because of reference problems.

PR #29, opened 2016-01-02
Add project to test MetadataExtractor.Portable

kwhopper:
Two date tests will still fail (originate from XmpCore). Directory.Tags needed to be IList for the Count property; IEnumerable doesn't have it.

### DIFF
--- a/MetadataExtractor.Tests/Formats/Exif/ExifDirectoryTest.cs
+++ b/MetadataExtractor.Tests/Formats/Exif/ExifDirectoryTest.cs
@@ -63,6 +63,7 @@ namespace MetadataExtractor.Tests.Formats.Exif
             JpegSegmentReader.ReadSegments(new SequentialByteArrayReader(thumbData));
         }
 
+#if !PORTABLE
         [Fact]
         public void TestWriteThumbnail()
         {
@@ -81,7 +82,7 @@ namespace MetadataExtractor.Tests.Formats.Exif
                 File.Delete(thumbnailFile);
             }
         }
-
+#endif
 //    public void TestContainsThumbnail()
 //    {
 //        ExifSubIFDDirectory exifDirectory = new ExifSubIFDDirectory();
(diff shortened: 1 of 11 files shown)


drewnoakes:
Thanks, this is definitely needed. Will merge this asap.

drewnoakes:
It's possible to use the `Count()` extension method instead of the `Count` property.

Also I discovered an approach that allows having a different `packages.config` file per `.csproj` file. I'll apply that technique here too.

Will pick this up again in a few hours as I have to head out.

kwhopper:
Ok, but the build errors are peppered all over IptcReaderTest and JfifReaderTest. Count doesn't exist, and unable to index IEnumerable. It seemed easier to just return IList instead of trying to fix them all, plus maintain compatibility with IReadOnlyList from non-portable. Thanks

kwhopper:
The DateTime test commit (95a78d1) worked on my machine, but I guess still didn't pass on appveyor. The rational for the change is it didn't seem like the second Assert in each function was a valid test. I think it would always be false -- if DateTime is created without specifying a DateTimeKind, it's set to Unspecified while the Xmp parsed datetime has a zone.

I could be entirely wrong here. DateTime is tricky business. It doesn't work for the Portable version (since it doesn't have TimeZone implemented completely) so I'm still going to investigate anyway. Pass along any thoughts you have. Thanks

drewnoakes:
The `MetadataExtractor.Portable.Tests` project is not, itself, a portable assembly. Is this intentional?

Will look at DateTime stuff once the portable test project is merged in. You're right, it's a tricky business and I don't have a clear idea off the top of my head.

drewnoakes:
So I looked into making the test assembly portable, and the first problem I saw is that there's no way to open a file from the file system! That means that none of the test data can be loaded. I'm not sure whether there are problems with having a non-PCL test project running PCL tests, but right now I don't see an alternative.

kwhopper:
Correct - MetadataExtractor.Portable.Tests isn't portable on purpose. It's only duty is to reference the portable MetadataExtractor assembly and run the same tests. I think you discovered why it can't be a portable project on its own.

The portable assembly is built against the PCL and isn't allowed to go outside it regardless of the program it's running in. Users of the portable assembly have to write for the specific requirements of their environment anyway, such as filesystem access. So, this seems reasonable to me. At least this gives us coverage of the current unit tests and helps identify any incompatibilities. Thanks!

drewnoakes:
Hehe, yes it seemed obvious in retrospect. It's great having a test project
for the PCL build. Thanks very much for making it happen.

I moved the XMP DateTime problem to a new issue for now.

Issue #19, opened 2015-06-28
Encoding problems in Canon makernote "Owner Name" string


drewnoakes:
```diff
diff --git a/jpg/metadata/canon ixus 750.jpg.txt b/jpg/metadata/canon ixus 750.jpg.txt
index 2fe3e71..55fcb6d 100644
--- a/jpg/metadata/canon ixus 750.jpg.txt	
+++ b/jpg/metadata/canon ixus 750.jpg.txt	
@@ -137,7 +137,7 @@ FILE: Canon IXUS 750.jpg
 [Canon Makernote - 0x0006] Image Type = IMG:DIGITAL IXUS 750 JPEG
 [Canon Makernote - 0x0007] Firmware Version = Firmware Version 1.00
 [Canon Makernote - 0x0008] Image Number = 1003892
-[Canon Makernote - 0x0009] Owner Name = Renè Glanzer
+[Canon Makernote - 0x0009] Owner Name = Ren� Glanzer
 [Canon Makernote - 0x000d] Camera Info Array = [94 values]
 [Canon Makernote - 0x0010] Canon Model ID = 26214400
```

```diff
diff --git a/jpg/metadata/canon powershot g11.jpg.txt b/jpg/metadata/canon powershot g11.jpg.txt
index 98eb3fb..5f23627 100644
--- a/jpg/metadata/canon powershot g11.jpg.txt	
+++ b/jpg/metadata/canon powershot g11.jpg.txt	
@@ -141,7 +141,7 @@ FILE: Canon PowerShot G11.jpg
 [Canon Makernote - 0x0006] Image Type = IMG:PowerShot G11 JPEG
 [Canon Makernote - 0x0007] Firmware Version = Firmware Version 1.00
 [Canon Makernote - 0x0008] Image Number = 4120954
-[Canon Makernote - 0x0009] Owner Name = Balázs Iván József +36304028290
+[Canon Makernote - 0x0009] Owner Name = Bal�zs Iv�n J�zsef +36304028290
 [Canon Makernote - 0x000d] Camera Info Array = [171 values]
 [Canon Makernote - 0x0010] Canon Model ID = 40894464
 [Canon Makernote - 0x0026] AF Info Array 2 = [48 values]
```

kwhopper:
I think this happens because of differences in your C# and Java version default encodings when reading these null terminated strings that have no set designated in the exif. It's impossible to know what encoding was intended.

[...]

PR #31, opened 2016-01-05
Hold original string byte arrays to allow variable encoding

kwhopper:
This is a work-in-progress to keep byte arrays available to handle encoding issues, like Issue #19. I wanted to get your thoughts before proceeding any further - on the right track, implementation discussion, etc. I threw in CanonMakernoteDescriptor.GetOwnerNameDescription() to possibly demonstrate your suggestion in the Issue.

It does have a happy side effect - XMP Extract now uses the original byte array directly instead of encoding first.

### DIFF
--- a/MetadataExtractor.Tests/Formats/Png/PngMetadataReaderTest.cs
+++ b/MetadataExtractor.Tests/Formats/Png/PngMetadataReaderTest.cs
@@ -75,7 +75,7 @@ namespace MetadataExtractor.Tests.Formats.Png
             Assert.NotNull(pairs);
             Assert.Equal(1, pairs.Count);
             Assert.Equal("Comment", pairs[0].Key);
-            Assert.Equal("Created with GIMP", pairs[0].Value);
+            Assert.Equal("Created with GIMP", pairs[0].Value.ToString());
         }
 
         private static string CreateTestString(int year, int month, int day, int hourOfDay, int minute, int second)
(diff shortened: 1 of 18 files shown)


kwhopper:
This push highlights a larger issue. There are many non-standard ways the Makerdata offset can be set, according to exiftool. I know of at least one image in the images database that has an incorrect offset and there could be many more.

(Sorry for the merge conflicts. I'm still not sure how to properly push commits with github yet)

drewnoakes:
There's some good stuff in these commits, though I haven't had the time to review them properly. The library definitely needs to store values in an unprocessed form internally, allowing custom interpretation on the way out (such as character encoding). I have been working towards this occasionally in the new-api branch. That's a larger unit of work though, so your PR might make a sensible stepping stone along the way. Whatever happens here should also be possible in the Java version -- just another thing to consider.

The second commit would probably be better suited to another PR as it's functionally unrelated, if I understand correctly. Anyway, if you want to make multiple PRs, it's a good idea to create a new branch for each one (rather than using your fork's master). That way you can develop each individually. Also, whenever you update that branch, the PR updates automatically.

So apologies for the delay in looking at this. You're tackling big issues, which is awesome, but requires more time to review than quick bug fixes.

kwhopper:
If it's ok with you, I'm going to close this and re-open it as branched PR's instead. Also, if I understand your explanation, I can continue making changes in other branches/PR's that won't cross into each other. I will also delete my current master branch (will this delete the PR code references?). There are some other updates I want to propose as separate PR's. The early mistake in creating this one on master is slowing those down. Thanks

PR #33, opened 2016-01-26
Add and display 13 new Canon tags

kwhopper:
Delved into Exiftool a bit to identify some Canon tags and offer some potential display cleanup for a few existing tags.

New/Revised Canon: TagRecordMode, TagMaxAperture, TagMinAperture, TagFocusContinuous, TagAESetting, TagDisplayAperture, TagZoomSourceWidth, TagZoomTargetWidth, TagSpotMeteringMode, TagPhotoEffect, TagManualFlashOutput, TagColorTone, TagSRAWQuality

A couple of functions were also converted from Exiftool to C# and I tried to attribute those appropriately.

### DIFF
--- a/MetadataExtractor/DirectoryExtensions.cs
+++ b/MetadataExtractor/DirectoryExtensions.cs
@@ -30,6 +30,44 @@ namespace MetadataExtractor
 {
     public static class DirectoryExtensions
     {
+        #region Int16
+
+        /// <summary>Returns a tag's value as a <see cref="short"/>, or throws if conversion is not possible.</summary>
+        /// <remarks>
+        /// If the value is <see cref="IConvertible"/>, then that interface is used for conversion of the value.
+        /// If the value is an array of <see cref="IConvertible"/> having length one, then the single item is converted.
+        /// </remarks>
+        /// <exception cref="MetadataException">No value exists for <paramref name="tagType"/>, or the value is not convertible to the requested type.</exception>
+        public static short GetInt16(this Directory directory, int tagType)
+        {
+            short value;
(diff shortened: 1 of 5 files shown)


drewnoakes:
Between now and the time I get to review this, perhaps you could try running it over the sample images?

Specifically, clone [this (large) repo](https://github.com/drewnoakes/metadata-extractor-images), then run `MetadataExtractor.Tools.FileProcessor.Program.Main2` (you have to rename it to `Main` -- sorry so hacky... the cmd line tools need some work) with argument `--text "<path>\metadata-extractor-images"`. Then run a `git diff` on the repo. You will see some changes in the `*\metadata\*.txt` files.

The sample images repo contains the output of the Java library. The .NET library currently produces some differences around text encoding and some time zone stuff. But if you run the above with your changes, you'll be able to see if it's introduced any other more significant changes to the output for any of the sample images. It's quite common that changes work on one camera model but introduce breaks on others. I'd have to run these checks before merging this PR, so it'd be smoother if you ran them yourself. The only problem is that the repo is quite large, but you only have to clone it once.

kwhopper:
Ok - I went through all the images per your instructions. There were no new errors that I could see; most files changed mod dates but had no textual changes. Text and datetimes were different, but that was expected. "Focal Plane" changes tended to show up the most outside Canon-model images.

Otherwise, I couldn't find anything new or that wasn't expected output. If there's something else to check let me know.

drewnoakes:
Thanks, I've merged this as 444f19c4938c2220ecf05c2e3f7b7a5be0a706d8. Nice work on those tricky AF tags. The documentation you added makes the strange logic clear.

PR #34, opened 2016-01-27
Hold original string byte arrays to allow variable encoding (second try)

kwhopper:
Replaces part of Issue #31, and some discussion is there. This puts it in another branch for now.

This is a work-in-progress to keep byte arrays available for user-defined encoding, like Issue #19. As a bonus, XMP Extract uses the original byte array directly instead of encoding the whole thing first.

### DIFF
--- a/MetadataExtractor/DirectoryExtensions.cs
+++ b/MetadataExtractor/DirectoryExtensions.cs
@@ -338,7 +338,7 @@ namespace MetadataExtractor
         }
 
         /// <summary>Gets the specified tag's value as an byte array, if possible.</summary>
-        /// <remarks>Only supported where the tag is set as String, Integer, int[], byte[] or Rational[].</remarks>
+        /// <remarks>Only supported where the tag is set as StringValue, String, Integer, int[], byte[] or Rational[].</remarks>
         /// <returns>the tag's value as a byte array</returns>
         [CanBeNull]
         public static byte[] GetByteArray(this Directory directory, int tagType)
@@ -347,6 +347,10 @@ namespace MetadataExtractor
 
             if (o == null)
                 return null;
+
+            var strvalue = o as StringValue;
+            if (strvalue != null)
+                return strvalue.Bytes;
 
             byte[] bytes;
 
--- a/MetadataExtractor/Formats/Png/PngMetadataReader.cs
+++ b/MetadataExtractor/Formats/Png/PngMetadataReader.cs
@@ -46,6 +46,9 @@ namespace MetadataExtractor.Formats.Png
     /// <author>Drew Noakes https://drewnoakes.com</author>
     public static class PngMetadataReader
     {
+        // by spec, PNG is generally supposed to use this encoding
+        private static string defaultEncodingName = "ISO-8859-1";
+
         private static readonly HashSet<PngChunkType> _desiredChunkTypes = new HashSet<PngChunkType>
         {
             PngChunkType.IHDR,
@@ -120,6 +123,11 @@ namespace MetadataExtractor.Formats.Png
         /// <exception cref="System.IO.IOException"/>
         private static IEnumerable<Directory> ProcessChunk([NotNull] PngChunk chunk)
         {
+            // For more guidance:
+            // http://www.w3.org/TR/PNG-Decoders.html#D.Text-chunk-processing
+            // http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.iCCP
+            var defaultEncoding = System.Text.Encoding.GetEncoding(defaultEncodingName);
+
             var chunkType = chunk.ChunkType;
             var bytes = chunk.Bytes;
 
@@ -179,7 +187,7 @@ namespace MetadataExtractor.Formats.Png
             else if (chunkType == PngChunkType.iCCP)
             {
                 var reader = new SequentialByteArrayReader(bytes);
-                var profileName = reader.GetNullTerminatedString(maxLengthBytes: 79);
+                var profileName = reader.GetNullTerminatedStringValue(maxLengthBytes: 79);
                 var directory = new PngDirectory(PngChunkType.iCCP);
                 directory.Set(PngDirectory.TagIccProfileName, profileName);
                 var compressionMethod = reader.GetSByte();
@@ -207,9 +215,11 @@ namespace MetadataExtractor.Formats.Png
             else if (chunkType == PngChunkType.tEXt)
             {
                 var reader = new SequentialByteArrayReader(bytes);
-                var keyword = reader.GetNullTerminatedString(maxLengthBytes: 79);
+                var keyword = reader.GetNullTerminatedStringValue(maxLengthBytes: 79).ToString(defaultEncoding);
                 var bytesLeft = bytes.Length - keyword.Length - 1;
-                var value = reader.GetNullTerminatedString(bytesLeft);
+                var value = reader.GetNullTerminatedStringValue(bytesLeft);
+                value.SetEncodingByName(defaultEncodingName);
+
                 var textPairs = new List<KeyValuePair> { new KeyValuePair(keyword, value) };
                 var directory = new PngDirectory(PngChunkType.iTXt);
                 directory.Set(PngDirectory.TagTextualData, textPairs);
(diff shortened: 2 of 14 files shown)


drewnoakes:
Thanks for the PR, and good that you've sorted out your branches. I'm unable to review these for the next few days but hope to get to this and your other work by the end of the week.

drewnoakes:
Looking through this, I like the general idea of storing unprocessed string data in `Directory` objects, so that the user can interrogate the raw data if they need to (usually related to encoding issues).

At first I thought we could just store `byte[]` for each string, but actually when pulling data out for tags which don't have dedicated descriptor methods, it's unclear whether to treat the data as text or not. So we need some kind of wrapper, as you have.

Also, in some cases the extracted `byte[]` can have a known/assumed encoding (as you pointed out with PNG).

So this raw string value should contain, a `byte[]` and an optional encoding, or possibly multiple encodings.

Then on the way out we can decode, or decode using a custom encoding.

The `new-api` branch is heading down a road that would provide this kind of thing quite elegantly, but that branch isn't ready for the prime time yet unfortunately.

So now that I've finally given this a proper reading through I've decided that I quite like it :) I'll pop a few comments inline and move towards merging it in after some testing. But not tonight sorry, it's too late here.

MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -207,9 +215,11 @@ private static IEnumerable<Directory> ProcessChunk([NotNull] PngChunk chunk)
             else if (chunkType == PngChunkType.tEXt)
             {
                 var reader = new SequentialByteArrayReader(bytes);
-                var keyword = reader.GetNullTerminatedString(maxLengthBytes: 79);
+                var keyword = reader.GetNullTerminatedStringValue(maxLengthBytes: 79).ToString(defaultEncoding);
                 var bytesLeft = bytes.Length - keyword.Length - 1;
```
drewnoakes: I think there's a bug here for encodings where `string.Length != bytes.Length` (multi-byte characters).

MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -207,9 +215,11 @@ private static IEnumerable<Directory> ProcessChunk([NotNull] PngChunk chunk)
             else if (chunkType == PngChunkType.tEXt)
             {
                 var reader = new SequentialByteArrayReader(bytes);
-                var keyword = reader.GetNullTerminatedString(maxLengthBytes: 79);
+                var keyword = reader.GetNullTerminatedStringValue(maxLengthBytes: 79).ToString(defaultEncoding);
```
drewnoakes: It'd be nice to pass the encoding into `GetNullTerminatedStringValue`, so that it is paired with the bytes. After, just call ToString. Doesn't matter so much for this anonymous temporary, but in general I think that's a nicer API. It makes you think about the encoding when you're reading the string out.

MetadataExtractor/Formats/Xmp/XmpReader.cs
```diff
@@ -130,7 +130,7 @@ public XmpDirectory Extract([NotNull] byte[] xmpBytes)
         /// <remarks>
         /// The extraction is done with Adobe's XMPCore library.
         /// </remarks>
-        public XmpDirectory Extract([NotNull] string xmpString) => Extract(Encoding.UTF8.GetBytes(xmpString));
+        //public XmpDirectory Extract([NotNull] string xmpString) => Extract(Encoding.UTF8.GetBytes(xmpString));
```
drewnoakes: In general I don't like to remove public APIs. Is there a good reason for this?

MetadataExtractor/IO/IndexedReader.cs
```diff
@@ -318,18 +318,39 @@ public string GetString(int index, int bytesRequested, [NotNull] Encoding encodi
...
+        /// <param name="index">The index within the buffer at which to start reading the string.</param>
+        /// <param name="maxLengthBytes">
+        /// The maximum number of bytes to read.  If a zero-byte is not reached within this limit,
+        /// reading will stop and the string will be truncated to this length.
+        /// </param>
+        /// <returns>The read <see cref="StringValue"/></returns>
+        /// <exception cref="System.IO.IOException">The buffer does not contain enough bytes to satisfy this request.</exception>
+        public StringValue GetNullTerminatedStringValue(int index, int maxLengthBytes)
+        {
             // NOTE currently only really suited to single-byte character strings
```
drewnoakes: I don't see why this comment applies. In fact I don't know that it was ever correct. Am I missing something?

MetadataExtractor/IO/IndexedReader.cs
```diff
@@ -318,18 +318,39 @@ public string GetString(int index, int bytesRequested, [NotNull] Encoding encodi
...
             var bytes = GetBytes(index, maxLengthBytes);
             // Count the number of non-null bytes
             var length = 0;
             while (length < bytes.Length && bytes[length] != 0)
                 length++;
-            return Encoding.UTF8.GetString(bytes, 0, length);
+
+            var _actualBytes = new byte[length];
+            if (length > 0)
+                Array.Copy(bytes, 0, _actualBytes, 0, length);
```
drewnoakes: Could be nice to avoid the copy here in the case that the arrays are the same length, as is done in `SequentialReader`. Actually there are two methods here that are very similar now.

MetadataExtractor/StringValue.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            get { return _buffer; }
+        }
+
+        public void SetEncodingByName(string encodingName)
+        {
+            _encoder = Encoding.GetEncoding(encodingName);
+        }
+
+        public Encoding Encoder
```
drewnoakes: `[CanBeNull]`

MetadataExtractor/StringValue.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            get { return _encoder; }
+            set { _encoder = value; }
+        }
+
+        public int Length
```
drewnoakes: Probably should clarify this as `ByteCount` as `Length` may be misinterpreted as the number of characters in the string, which won't always be the same thing.

MetadataExtractor/DirectoryExtensions.cs
```diff
@@ -348,6 +348,10 @@ public static byte[] GetByteArray(this Directory directory, int tagType)
             if (o == null)
                 return null;
 
+            var strvalue = o as StringValue;
+            if (strvalue != null)
+                return strvalue.Bytes;
+
             byte[] bytes;
 
             var rationals = o as Rational[];
```
drewnoakes: Shouldn't a few other methods in `DirectoryExtensions` become `StringValue`-aware? At the very least, `GetString` should, with an optional encoding argument.

Maybe also `DirectoryExtensions.GetStringValue`, though I don't think that should do any coercion of types as going from, say, a double, to a string, to a byte[] seems pretty dumb and error prone.

MetadataExtractor/StringValue.cs
```diff
@@ -0,0 +1,172 @@
...
+        public string ToString(Encoding encoder)
+        {
+            return encoder.GetString(_buffer, 0, Length);
+        }
+
+        #endregion
+
+        // user-defined conversion from StringValue to Double
+        /*public static implicit operator double(StringValue strval)
+        {
```
drewnoakes: `implicit` conversions often feel like they'd be nice, but they generally hide too much magic and make things hard to debug. I agree with having these commented (or removed).

MetadataExtractor/Formats/Png/PngMetadataReader.cs
kwhopper: Yes, probably (without testing it). In the previous code, it only ever returned a string so there was nothing else to go on. StringValue has a Length that counts the bytes, which is better anyway:

`var keywordStringValue = reader.GetNullTerminatedStringValue(maxLengthBytes: 79);`
`var bytesLeft = bytes.Length - keywordStringValue.Length - 1;`

Subsequent places after that use `keyword` could use `keywordStringValue.ToString([some encoding])` instead

(untested on my end)

MetadataExtractor/Formats/Png/PngMetadataReader.cs
kwhopper: I went back and forth on this one originally, and agree - it's a toss-up. You could certainly pass it along, but in the end it seemed like the reader should stick to 'reading' and let the user decide later how it should be interpreted explicitly. An overload could accomplish it in the near term; if it was required up-front and the default encoding for that reader always matched, it would seem a little redundant. But again, this feels like a programmer 'taste' issue.

MetadataExtractor/Formats/Xmp/XmpReader.cs
kwhopper: Again, I debated this as well so left it in there commented-out. As StringValue is used to hold the original XMP bytes from a file, the byte[] overload of this method (`public XmpDirectory Extract([NotNull] byte[] xmpBytes)`) seemed to accomplish the internal requirements. Before this, the bytes were converted to UTF-8 strings when read, and then passed to this function - which turned them back into bytes. A version that assumes a certain encoding could be prone to issues - even from a string supplied by the user.

Correct me if I'm wrong - the XMP spec indicates UTF-8 is preferred. However, other encodings are accepted. Can you know for sure that `string xmpString` is UTF-8? Perhaps if this version accepted an encoding input it might still be safe, something like
`public XmpDirectory Extract([NotNull] string xmpString, Encoding encoder) => Extract(encoder.GetBytes(xmpString));`
I can see why users might call this directly with their own XMP string (probably so XmpCore can parse it for them) but imposing a UTF-8 encoding here doesn't seem like a good idea.

MetadataExtractor/IO/IndexedReader.cs
kwhopper: Agreed, since it's now just grabbing bytes. The previous code might have just had the comment out of place. The call `return Encoding.UTF8.GetString(bytes, 0, length);` in the old code might have been the intended line. Whether the comment even described that issue correctly, I'm not sure but it doesn't seem like it needs to be there anymore.

MetadataExtractor/IO/IndexedReader.cs
kwhopper: Yeah, that could work. Initially, I tried to do the least amount of code damage since the target was creating/reading a StringValue. The difference is SequentialReader uses a GetByte() loop while IndexedReader is using GetBytes(). I left those as they were.

You pretty much have to use the GetByte() loop version if you don't want to copy at the end; GetBytes() won't know it should stop at null bytes. Is there any appreciable performance hit with a single-byte loop?

MetadataExtractor/StringValue.cs
kwhopper: Agreed

MetadataExtractor/DirectoryExtensions.cs
kwhopper: I might leave that decision to you. Your experience with how people use these methods far exceeds mine. But, it seems like the two overloads of GetString() could be combined into one if you wanted to really enforce this. That would at least shorten the number of overloads.

I'm not so sure on a GetStringValue version -- unless it ended up calling a combined GetString() somehow. I think you might answer your own question there at the end; it doesn't seem to make much sense to hold a double in a byte[] in a class with 'String' in its name.

MetadataExtractor/StringValue.cs
kwhopper: Yep. I tested them out, but came to a similar conclusion. It seemed a shame to just delete them, but removing is also fine.

kwhopper:
Did you want me to revise it, or were you going to use feedback to polish it up? Thanks for looking at this again.

drewnoakes:
Thanks for responding to all the comments. Very useful. I'm generally really happy with this and will merge it once I've had a chance to run it over the test images for any differences in output or performance. Thanks for your patience.

drewnoakes:
I finally found an hour to look at this on a train today. It's pushed (as 2f934658993384f255e88966768aeae1c283303e because I rebased) and I made a few tweaks along the lines discussed above, including fixing a compilation bug and a unit test failure. Also made `StringValue` a value type. It works great and, after implementing all the `IConvertible` methods, there are no regressions over the image database.

Thanks again for the contribute and your patience!

kwhopper:
This is great! It feels like a more natural way of handling this. Once it's on the Java side, I think a lot of little end-user issues will go away. Thanks!

drewnoakes:
Finally ported this to the Java version as well.

Issue #35, opened 2016-02-02
Reported JPEG Thumbnail Offsets are incorrect


kwhopper:
JPEG byte offsets for embedded thumbnails are not correct. The offset is calculated relative to the segment, and does not include the offset of the segment relative to the file start. Usually, it's off by the SOI + TIFF markers or 6 + 6 = 12. This would not be true for App1 fragments in the test data since they won't contain SOI, and makes this more complicated than it seems.

kwhopper:
I commented more in the referenced pull request. The fix in the PR may or may not be desirable.

twest820:
Drew, would it be possible to get @kwhopper's pull request into the next release?

In the meantime, I'm thinking to work around this using an approach similar to that illustrated in the spike code below.

            ExifThumbnailDirectory thumbnails = ...
            Assert.IsTrue(thumbnails.GetDescription(ExifThumbnailDirectory.TagCompression).StartsWith("JPEG"));
            long thumbnailOffset = Int64.Parse(thumbnails.GetDescription(ExifThumbnailDirectory.TagThumbnailOffset).Split(' ')[0]);
            const int maxIssue35Offset = 12;
            int thumbnailLength = Int32.Parse(thumbnails.GetDescription(ExifThumbnailDirectory.TagThumbnailLength).Split(' ')[0]) + maxIssue35Offset;
            byte[] thumbnail = new byte[thumbnailLength];
            using (FileStream imageStream = new FileStream(imagePath, FileMode.Open, FileAccess.Read, FileShare.Read))
            {
                imageStream.Seek(thumbnailOffset, SeekOrigin.Begin);
                imageStream.Read(thumbnail, 0, thumbnailLength);
            }

            // work around Metadata Extractor issue #35
            Assert.IsTrue(thumbnailLength > maxIssue35Offset + 1);
            int issue35Offset = 0;
            for (int offset = 0; offset <= maxIssue35Offset; ++offset)
            {
                // 0xffd8 is the JFIF start of image segment indicator
                if ((thumbnail[offset] == 0xff) && (thumbnail[offset + 1] == 0xd8))
                {
                    issue35Offset = offset;
                    break;
                }
            }

            using (MemoryStream thumbnailStream = new MemoryStream(thumbnail, issue35Offset, thumbnailLength - issue35Offset, false))
            {
[...]

PR #36, opened 2016-02-02
Attempt to hold onto absolute position of a Directory relative to the original file/stream

kwhopper:
This is the long way around to try and address Issue #35. It introduces a "JpegSegment" object to pass around instead of a byte[] for segments. The object contains the byte[] previously sent, plus a starting position from when the segment was read relative to original file.

In one sense, it would have been easier to correct the thumbnail offset directly in ExifTiffHandler.Completed. However, as I was going through this it seemed like there might be use down the road for the original offset. One thing lead to another and a bunch of files were changed. This kind of demonstrates how complex it is to send this around the current readers/handlers, so I'll leave it up to you whether it's a "good" thing. It does fix the thumbnail issue, but at a cost. That said, it's a totally optional property and isn't used anywhere else. Also, not all file types were tested so this might be incomplete.

### DIFF
--- a/MetadataExtractor.Tests/Formats/Exif/ExifReaderTest.cs
+++ b/MetadataExtractor.Tests/Formats/Exif/ExifReaderTest.cs
@@ -40,7 +40,7 @@ namespace MetadataExtractor.Tests.Formats.Exif
         [NotNull]
         public static IList<Directory> ProcessSegmentBytes([NotNull] string filePath)
         {
-            return new ExifReader().ReadJpegSegments(new[] { File.ReadAllBytes(filePath) }, JpegSegmentType.App1).ToList();
+            return new ExifReader().ReadJpegSegments(new[] { new JpegSegment(JpegSegmentType.App1, File.ReadAllBytes(filePath)) }, JpegSegmentType.App1).ToList();
         }
 
         [NotNull]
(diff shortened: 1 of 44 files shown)


drewnoakes:
I know it's been an age, but I recently started revisiting JPEG segment handling and I'll definitely get the spirit of this PR into the code somehow. Sorry again that it's taken so long to get to it.

drewnoakes:
I've been rebasing this to address the merge conflicts and doing some review.

In general we definitely stand to benefit from keeping track of some offsets during parsing.

Adding `StartPosition` to `Directory` is a great start, but it doesn't cover all cases.

Some directories don't really have a position:

- `FileMetadataDirectory` which models information from the file system)

Other directories potentially have multiple positions of interest, such as makernotes:

- JPEG segment offset
- TIFF data offset
- IFD offset

I agree that these values should be stored in properties on the directory classes themselves. But given the variations, modelling it `Directory` subclasses might be more versatile than having a single offset on  the base.

(I'm still going through the changes, to ideas may vary in coming days. I'm capturing thoughts here as I go, as I only have short bursts of time to look at this now.)

kwhopper:
Closing as #130 makes this obsolete. The general idea can be implemented much more easily if #130 is integrated first.

PR #37, opened 2016-02-08
Olympussubifd

kwhopper:
Olympus (and maybe others) can have subifd pointers (data format 13) stored in normal tags. This commits tries to properly handle these by not popping the last Directory when processing. Fixes a number of Olympus files in the images repo that previously stopped processing with format 13.


kwhopper:
The first push sort of worked, but it was adding all subifd tags to the parent ifd. This created overlap when the tagId's were the same and values were overwritten. Since subifd's are really Directories in their own right, this adds a way to allow the parent to register possible child ifd's in a lookup table and store them in separate directories.

By definition, this requires creating more Directory classes. First - want your thoughts on this idea. Second - if it needs classes defined in cs files, how best to organize those? For example, Olympus has at least 7 subifd's.

Lastly, I'm still looking at it but I think "CameraSettings" in Olympus is really a subifd. If this works out, it might be possible to convert it over. I'm not 100% sure on that though.

kwhopper:
All Equipment tags are in there now

drewnoakes:
This is looking good!

You've hit the nail on the head regarding the explosion of code required for new directories. The design that's used is definitely sub-optimal, and harks back to the very early days of my development career. There is a `new-api` branch in which I'm exploring what I think will be a much better way of modelling known tags and their decoding. For example, [here's how the base Exif tags would be defined](https://github.com/drewnoakes/metadata-extractor-dotnet/blob/new-api/MetadataExtractor/ExifTags.cs). The name, id, expected type and descriptor function are all co-located in one place rather than spread out across multiple classes. It's early days for this work, and I'm not sure it'll be easy to back-port these changes to the Java version, but that's the direction I think things should go. If you're interested, take a look at the commits on that branch.

I'd also like to preserve the tree structure of directories in the output of `ExifReader` and indeed all the other metadata readers. It could still be flattened into a single enumerable.

Your PRs are all awesome and I will sit down to review them properly asap. I'm pretty overloaded at the moment with two clients and a newborn, so thanks for your patience. If they were simple bug fixes, I could just eyeball them, but these are deep issues you're working on and they require some thought, especially as I'm in the throes of migrating the representation of tags and the modelling of extracted metadata to a completely new model. It's early days there, but if you are interested in reviewing the `new-api` branch, your thoughts would be appreciated. [This file](https://github.com/drewnoakes/metadata-extractor-dotnet/blob/new-api/MetadataExtractor/NewApi.cs) contains some half-baked sketches of how the TIFF parsing might hang together in future.

kwhopper:
No need to apologize at all. This is all volunteer work, and something new / interesting to tinker with. It sounds like the goal is a writable library, so I figured why not start the learning by fixing things that didn't work. I'll definitely look at your links and offer any ideas that come to mind. Thanks

drewnoakes:
Glad you're having fun. I definitely want to respect the time you're spending by reviewing and integrating your changes.

Yes, a goal is to support writing metadata. Getting a better model in place for doing so is a key part of it.

I'm pretty excited to see how these non-standard IFD pointers (format type 13) work across sample images.

Incidentally, if you find some good sample images and have that repo already cloned, feel free to add them. Ideally sample images should be straight from the camera (if advertised as of the camera model) or labelled by the software that produced/manipulated them. One of these days I'll walk into a camera store with an SD card and convince them to let me take a photo on each and every camera in the shop :)

kwhopper:
Again using ExifTool as a guide, this adds the 'Camera Settings' subifd for Olympus.

I ran this code over images repository and there weren't any obvious issues. You're free to verify that. What are your thoughts on integrating this branch? I'm happy to continue converting these but want to make sure it's worth the time. It could also be integrated and 'disabled' by returning false from ExifTiffHandler.IsTagSubIfdPointer manually - effectively ignoring this processing.

I'd also like to go fix some issues reported on the Java version, but do them in .NET first. The Java repo is working on my machine but it takes quite a bit longer to build with NetBeans and it really slows down dev. My .NET fork is so messy from the other branches that I'm worried I'll (accidentally) lose some of my branches. Thanks

kwhopper:
The last commit mostly fixes issue #151 on the Java version. It looks like some IFD pointers are using Tiff format 13 now, even in ExifIFD0.

kwhopper:
I noticed IsTagSubIfdPointer could accomplish what IsTagIfdPointer was doing without the if/else logic. This registers (Directory.RegisterSubIfd) the three IFD's checked in that method, which makes IsTagIfdPointer unnecessary.

Sorry if this branch seems to be going on and on. This last one is more of an optimization anyway, although I personally like where the 'register' part of this went. Simplified and stabilized subifd handling quite a bit.

drewnoakes:
I'm working through this PR now. It's great to see so many of the warnings about unsupported TIFF tag types go away and there's a lot more information available to Olympus users.

I squashed your commits into a single commit 01925e881ca13 (as there was a bit of back and forth over time) and then added another commit 4d5701027c with a few changes. Broadly:

- `Directory` is a generic type and shouldn't really know anything about TIFF IFDs. I've reverted back to having a lookup in the `ExifTiffHandler` for now.
- TIFF tag type 13 is still invalid according to the spec, so rather than add it to the enum, I've added an escape hatch (`TryCustomProcessFormat`) for adding logic that deals with unexpected codes. It achieves the same thing in the end, but keeps a clear division between the spec, and extensions to the spec.
- I added you as the `<author>` of the new directories of course!

Again I'm sorry that it took so long to get to this. I'm on holiday at the moment so have a bit of free time. This PR is a great addition to the project.

kwhopper:
Great! I'm glad you went through and cleaned it up, and no worries on the taking your time. There are still a few more of these to add for Olympus and I'll use your code corrections so there won't be so many things to fix.

Thanks for letting me contribute. It's good clean fun, nerdy as it is.  :)

PR #38, opened 2016-02-13
Describe ExifDescriptorBase.TagLensSpecification to match Java

kwhopper:
Completes to-do in commit ea3d8dc 

Tested against images database and couldn't find any differences. You should verify.


drewnoakes:
Nice!

PR #42, opened 2016-03-23
support reading up to two subifd's for tag 0x14a

kwhopper:
One way to fix drewnoakes/metadata-extractor#103 reported on the Java version.

NEF puts one or more (usually one or two) subifd's in tag 0x014a as either an IFD or one or more longs. This push supports reading them as longs, and only the first two. I don't consider this a complete solution, but it gets it started and will only run when that tag exists.

Two new Directory classes were added to support this. Review that decision, although class explosion is a well known issue for subifd's anyway.

### DIFF
--- a/MetadataExtractor/Formats/Exif/ExifTiffHandler.cs
+++ b/MetadataExtractor/Formats/Exif/ExifTiffHandler.cs
@@ -83,6 +83,12 @@ namespace MetadataExtractor.Formats.Exif
                 PushDirectory(typeof(ExifInteropDirectory));
                 return true;
             }
+
+            /*if(tagType == ExifIfd0Directory.TagSubIfdOffset && CurrentDirectory is ExifIfd0Directory)
+            {
+                return true;
+            }*/
+
             return false;
         }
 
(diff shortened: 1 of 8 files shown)


drewnoakes:
Firstly, I'm quite surprised that `ExifDirectoryBase.TagSubIfdOffset` (`0x014a`) was completely unused by the lib!

Secondly, it's interesting to think that there could be more than one pointer.

Some preliminary research shows that this tag is used in IFD0 (by both Sony ARW and Leica DNG) and in the thumbnail IFD (by Samsung SRW). This brings up the issue that currently there's no way of differentiating which image this sub-IFD applies to. I'll have a think about modelling hierarchy in the returned directories so that consumers can differentiaate. I think the current `ITiffHandler` infrastructure should allow this easily enough.

kwhopper:
> This brings up the issue that currently there's no way of differentiating which image this sub-IFD applies to.

Yeah, I think this has/will become a bigger issue as more subifd's are added. The class "explosion" problem is one thing, but like in this case, some hierarchy will be needed. The part that wasn't kept in #37 (where it tried to register subifd's in an ifd) indirectly tried to do this. That wasn't the initial goal but now that I think about it, it sort of imperfectly did part of what you're describing. Maybe register other directories instead of tags? OTOH, this might be a new-version issue; it could change the library quite a bit.

drewnoakes:
I think fcfb8d0 and aed57bc address the issue highlighted by this PR (excluding the hierarchy bit). You can see the significant impact it had on sample raw files in the image library [here](https://github.com/drewnoakes/metadata-extractor-images/commit/32df1ae9919d70946a5a47013da0b170069b65f7). Very cool!

Please review these changes if you have the chance.

kwhopper:
I think that works. Line 220 of TiffReader could be less cryptic by just checking the component size (your math is effectively doing that anyway)...

if(format.ComponentSizeBytes == 4)

Good job! I'll go through some of the unknown tags that came from this; there weren't too many.

drewnoakes:
I think it is possible (however unlikely) that 'format' is null (for
example if the type id is 13).

It'd be great to decode more of these unknown TIFF fields.

drewnoakes:
This has been ported to Java too now. Closing this. Reopen if you feel it's needed.

Issue #30, opened 2016-01-03
Reinstate XMP DateTime tests


drewnoakes:
As discussed in #29, the XMP tests related to DateTime values are problematic.

Both `TestExtract_OriginalDateTime` and `TestExtract_DigitizedDateTime` tests have been skipped, and should be reinstated and corrected.

It seems that maybe the `DateTimeOffset` type would be a better fit here. That may require changes to the `XmpCore` project as well.

drewnoakes:
Obsolete.

PR #43, opened 2016-03-29
Parse with DateTimeOffset instead of DateTime

kwhopper:
Date/Time parsing switched to DateTimeOffset. XMP date fields have always been returned as strings so no changes were necessary to the XMP project (that will come later). If you wanted to return these fields directly as some form of date object, that's a different issue.

The two XMP tests were reinstated to satisfy #30. Two new parsing formats were added to DirectoryExtensions to make those tests work. For now, it has to be that way because the XMP fields are still strings.

### DIFF
--- a/MetadataExtractor.Tests/DirectoryTest.cs
+++ b/MetadataExtractor.Tests/DirectoryTest.cs
@@ -114,27 +114,25 @@ namespace MetadataExtractor.Tests
         [Fact]
         public void TestSetStringAndGetDate()
         {
-            Action<string, DateTime> test = (str, expected) =>
+            Action<string, DateTimeOffset> test = (str, expected) =>
             {
                 _directory.Set(1, str);
-                Assert.Equal(expected, _directory.GetDateTime(1));
+                Assert.Equal(expected, _directory.GetDateTimeOffset(1));
             };
 
-            // TODO revisit these commented cases and introduce GetDateTimeOffset impl/test
-
-            test("2002:01:30 23:59:59",           new DateTime(2002, 1, 30, 23, 59, 59,     DateTimeKind.Unspecified));
-            test("2002:01:30 23:59",              new DateTime(2002, 1, 30, 23, 59,  0,     DateTimeKind.Unspecified));
-            test("2002-01-30 23:59:59",           new DateTime(2002, 1, 30, 23, 59, 59,     DateTimeKind.Unspecified));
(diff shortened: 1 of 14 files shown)


drewnoakes:
This is interesting. I'm not sure that `DateTimeOffset` is the right choice in all cases though. Setting the offfset to zero indicated UTC/GMT, but sometimes the offset is simply unknown.

I'm living in GMT for half the year, but recently was in GMT+9 which really brought the point home for me as I worked on some of these date issues. If a date value has offset of zero, then when I displayed it on my machine it would be adjusted to show as 9 hours earlier.

Sometimes we just don't have an offset. Sometimes we do. Unfortunately .NET doesn't have a single type that reflects this. Instead we have `DateTimeOffset` for the former case and `DateTime` for the latter. There's no `DateTimeWithNullableOffset` or equivalent with less verbose name.

The inverse problem exists in the Java implementation, where the `java.util.Date` class always has an offset. In that way, there's no reliable way of exposing the metadata's value reliably.

In the Java implementation, recent commits encourage storing dates as strings (which is often the case when they're read from files anyway) and parsing them on the way out according to the user's specified time zone. This seems like a sensible approach, and definitely works better than what was there before.

On the .NET side, I'm less sure of what works reliably. I will give this PR a more thorough look as soon as possible, but wanted to say the above and see if you have any thoughts related to it before I do review it.

Cheers!

kwhopper:
Don't worry - I'm still struggling a little with this one. The problem I see is you're always parsing a stored string which makes the DateTime class pretty unreliable. It will tend to default to your current time zone, as you explained. From my testing, it always seemed like I had to correct for that in arbitrary ways. TimeZoneInfo was ok (like in XmpCore) but since Portable doesn't have a TimeZoneInfo class...

If the stored string of a date doesn't have an offset, is there any other way to make it portable than assuming it's UTC? Does this PR come across to you as always assuming a 0 offset, or was there a particular test case that seemed wrong? What kind of Date represented as a string would be "incorrect" from your perspective?

You're certainly more familiar with metadata date/time issues than I, but this one seems like a pretty big deal to define moving forward. I'm happy to continue the discussion until we both understand it.  :)

kwhopper:
... and interestingly (if I did it right) there were no diffs in the Images project when I ran FileProcessor after this change. I was a bit surprised, so after that I decided to proceed with this PR. It's certainly possible that the Images database doesn't present all the possible problems with this type of parsing change.

PR #44, opened 2016-03-30
Olympusrawdev

kwhopper:
Implements another Olympus subifd (Raw Development, 0x2030) as a directory. Cleans-up more format 13 errors from the images project.


drewnoakes:
Looks great.

Before I merge it, can you remove the first two commits just to keep things clearer? You can do that via an interactive rebase. Let me know if it's something you'd like some help with. Also, I've fetched your commits to my machine as a backup, so don't worry about losing anything.

kwhopper:
That was... odd, but I think it's ok now. If not, let me know.

drewnoakes:
Perfect. The build failed due to an error on the AppVeyor side it seems. Have triggered a new build and will merge if/when it passes.

PR #46, opened 2016-03-31
Olympus Image Processing subifd (0x2040) directory

kwhopper:
Implements another Olympus subifd (Image Processing, 0x2040) as a directory. Cleans-up a good number of format 13 errors from the images project.


drewnoakes:
Nice!

PR #47, opened 2016-04-02
Olympus Focus Info subifd (0x2050) directory

kwhopper:
Implements an Olympus subifd (Focus Info, 0x2050) as a directory. Cleans-up the remaining format 13 errors from the images project. Other notes:

- There are two "Serial Number" tags in the main Olympus Makernote. This corrects the list and reassigns TagFirmware to value 0x0405

- FocusInfo tag 0x4000 is a subifd pointing to a second OlympusMakernoteDirectory. Added that to ExifTiffHandler.TryEnterSubIfd. That corrects a few other format 13 errors as well.


drewnoakes:
Good stuff.

PR #55, opened 2016-10-02
Added support to PngMetadataReader for zTXt chunks.

Baloogan:


drewnoakes:
Very nice. Do you have a sample image that includes a `zTXt` chunk? There aren't any in the sample image repo, and it'd be useful to have one for testing.

drewnoakes:
I found some test images here http://www.schaik.com/pngsuite/

Issue #23, opened 2015-09-06
Can this edit metadata?


urza:
Hi, this looks fantastic. Can it also edit and save metadata to files? Or just read? If not, can you recommend any c# library for writing metadata? (description, comments, tags)
Thnaks

drewnoakes:
No support for writing metadata at this stage. Hopefully one day, but there
is no ETA for this.

As for other options I can not make any recommendations. If you find
anything, please do report back on this issue. Thanks!

RowlandShaw:
One option could be to write back to an XMP file, and perhaps add support to read from the XMP semi-automatically (i.e. where platforms support interrogating the file system, and the method overload allows)?

michaelosthege:
Hi folks,
I have a working implementation for writing XMP metadata. There are a few things required in [XmpCore](https://github.com/drewnoakes/xmp-core-dotnet) to, so I am creating several pull requests...

1. [PR15 on XmpCore](https://github.com/drewnoakes/xmp-core-dotnet/pull/15) > addition of two methods to XmpCore.XMPMetaFactory and XmpCore.Impl.XmpMetaParser
2. PR #64 > addition of XDocument Root property to MetadataExtractor.Formats.Xmp.XmpDirectory and two lines in MetadataExtractor.Formats.Xmp.XmpReader.Extract
3. upcoming PR > addition of MetadataExtractor.ImageMetadataWriter and MetadataExtractor.Formats.Jpeg.JpegMetadataSubstitutor for substituting XMP in existing Jpeg files

Please note that my implementation works at the level of the XDocument.

I'm looking forward to depend my project on the NuGet packages again when all of this is merged...

cheers,
Michael

Selzier:
Is there any updates on PR #65 ? I also need to write (or rather delete) some XMP metadata, and I have not been able to achieve this. In Python with libxmp, I can do this:
```
if xmp.does_property_exist(ns, u'property'):
xmp.delete_property(XMP_NS_GPHOTOS_AUDIO, u'GAudio:Mime')

# save stripped XMP header to original image
if xmpfile.can_put_xmp(xmp):
xmpfile.put_xmp(xmp)
xmpfile.close_file()
```

Is this possible with XMPCore? How do you access the metadata as a xdoc? I've tried this but is failing:
```
XDocument file = XDocument.Load(filename);
var xmpMeta = XmpMetaFactory.ParseFromXDocument(file);
```

[...]

PR #65, opened 2016-10-24
added XmpDirectory.Root and classes for writing Xmp to Jpeg files

michaelosthege:
Adds a Root property to XmpDirectory such that the XDocument can be accessed directly.
Load behavior is modified to set the Root property.
Writing is accomplished by substituting existing Xml in App1 segments with new Xml, or inserting it if not present in the first place.

See #23 for more information.

### DIFF
--- a/MetadataExtractor/Formats/Jpeg/JpegMetadataWriter.cs
+++ b/MetadataExtractor/Formats/Jpeg/JpegMetadataWriter.cs
@@ -1,0 +1,218 @@
+#region License
+//
+// Copyright 2002-2016 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+// More information about this project is available at:
+//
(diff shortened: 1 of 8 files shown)


drewnoakes:
Just wanted to say that I'm pumped to get this feature into the library, but I want to give it a proper review because it'll be the first time the library is writing data and therefore the first time the library could potentially destroy users' files. I'm really busy at the moment but hope to find some time for this soon. Thanks for your patience.

drewnoakes:
Quick update. I've reworked the JPEG _reading_ code to be more robust. It will allow better performance and safety for injecting/replacing XMP in JPEG files.  Haven't pushed it yet as I want to do some more tidying and it's late here now.

michaelosthege:
I just updated my branch with your latest changes & also some refactorings.
+ refactored it such that the algorithm is easier to understand, review & test,
+ Unit tests for the Jpeg Xmp writing are still ToDo - any suggestions for a good test case?

I had used `http://ns.adobe.com/xap/1.0/\x0` as a segment preamble before and now I use `XmpReder.JpegSegmentPreamble` wich equals `http://ns.adobe.com/xap/1.0/\0` (without the x).  (My reference was [this](http://dev.exiv2.org/projects/exiv2/wiki/The_Metadata_in_JPEG_files))

In both reading & [writing](https://github.com/drewnoakes/metadata-extractor-dotnet/pull/65/files#diff-cf437b89164f79344877a2490237b79fR154), **could the differences in the end of the preamble lead to segments that are not recognized as XMP segments?** (And thereby ignoring them in read, or adding a duplicate segment in write?)

elangelo:
@drewnoakes any chance this can be picked up again? i'm willing to merge the conflicts and to follow up on any remarks you still have after review.

drewnoakes:
@elangelo the barriers to merging this are 1) my free time to review and really think deeply about this major change to the library and 2) my reluctance to take on the risk and responsibility involved with writing metadata to files and potentially breaking people's data. The cost of bugs becomes _much_ higher once we start writing data. That said, there's regular interest in doing so. If you're interested in helping with this PR, then reviewing the general approach and thinking through how other kinds of metadata may be written would be helpful. Thanks.

Issue #61, opened 2016-10-07
Remove calls to Debug.WriteLine


drewnoakes:
There are a few of these calls throughout the library. Review each case, and remove them.

drewnoakes:
Closed by @kwhopper in #68.

PR #68, opened 2016-11-07
Remove calls to Debug.WriteLine #61

kwhopper:
Attempt to address #61

### DIFF
--- a/MetadataExtractor/Formats/Exif/ExifReader.cs
+++ b/MetadataExtractor/Formats/Exif/ExifReader.cs
@@ -75,20 +75,21 @@ namespace MetadataExtractor.Formats.Exif
         {
             var directories = new List<Directory>();
 
+            var exifTiffHandler = new ExifTiffHandler(directories);
             try
             {
                 // Read the TIFF-formatted Exif data
-                TiffReader.ProcessTiff(reader, new ExifTiffHandler(directories));
+                TiffReader.ProcessTiff(reader, exifTiffHandler);
             }
             catch (TiffProcessingException e)
             {
                 // TODO what do to with this error state?
-                Debug.WriteLine(e);
+                exifTiffHandler.Error("ExifReader TiffProcessingException : " + e.Message);
             }
             catch (IOException e)
             {
                 // TODO what do to with this error state?
-                Debug.WriteLine(e);
+                exifTiffHandler.Error("ExifReader IOException : " + e.Message);
             }
 
             return directories;
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusMakernoteDirectory.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusMakernoteDirectory.cs
@@ -441,16 +441,8 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             var reader = new SequentialByteArrayReader(bytes);
             var count = bytes.Length / 4;
 
-            try
-            {
-                for (var i = 0; i < count; i++)
-                    Set(CameraSettings.Offset + i, reader.GetInt32());
-            }
-            catch (IOException e)
-            {
-                // Should never happen, given that we check the length of the bytes beforehand.
-                Debug.WriteLine(e);
-            }
+            for (var i = 0; i < count; i++)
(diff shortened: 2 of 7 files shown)


MetadataExtractor/Formats/Exif/makernotes/OlympusMakernoteDirectory.cs
```diff
@@ -441,16 +441,8 @@ private void ProcessCameraSettings([NotNull] byte[] bytes)
...
-            catch (IOException e)
-            {
-                // Should never happen, given that we check the length of the bytes beforehand.
-                Debug.WriteLine(e);
-            }
+            for (var i = 0; i < count; i++)
+                Set(CameraSettings.Offset + i, reader.GetInt32());
         }
 
         public bool IsIntervalMode()
```
kwhopper: I agree based on the 'count' math above - this error should never happen

MetadataExtractor/Formats/Png/PngHeader.cs
```diff
@@ -43,10 +43,7 @@ public PngHeader([NotNull] byte[] bytes)
             BitsPerSample = reader.GetByte();
 
             var colorTypeNumber = reader.GetByte();
-            var colorType = PngColorType.FromNumericValue(colorTypeNumber);
-            if (colorType == null)
-                throw new PngProcessingException("Unexpected PNG color type: " + colorTypeNumber);
-            ColorType = colorType;
+            ColorType = PngColorType.FromNumericValue(colorTypeNumber);
```
kwhopper: With "Unknown" available, colorType will never be null

MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -102,14 +102,7 @@ public static
...
             {
-                try
-                {
-                    directories.AddRange(ProcessChunk(chunk));
-                }
-                catch (Exception e)
-                {
-                    Debug.WriteLine(e);
-                }
+                directories.AddRange(ProcessChunk(chunk));
```
kwhopper: There are various problems that can happen in ProcessChunk, so my vote is to let the exception happen and try to fix it afterwards.

MetadataExtractor/Formats/WebP/WebpRiffHandler.cs
```diff
@@ -165,17 +190,23 @@ public void ProcessChunk(string fourCc, byte[] payload)
...
+                    {
+                        directory.Set(WebPDirectory.TagImageWidth, width);
+                        directory.Set(WebPDirectory.TagImageHeight, height);
                     }
+                    else
+                        directory.AddError(error);
+                    _directories.Add(directory);
                     break;
                 }
             }
```
kwhopper: For all of the cases in WebRiffHandler, the replacement goal was to always return a directory. If data isn't valid, those Sets are avoided but AddError will ensure the problem is propagated for possible review.

kwhopper:
With "Unknown" available for Png, a few of the Png files from metatdata-extractor-images will emit more data instead of failing. This changes the images output a bit so it might be good to check over those and see if they seem appropriate. It's also possible other output may change with the WriteLine replacements.

MetadataExtractor/Formats/Exif/makernotes/OlympusMakernoteDirectory.cs
drewnoakes: Agree. This is likely a remnant from the Java version where checked exceptions can't determine this kind of thing.

MetadataExtractor/Formats/Exif/ExifReader.cs
```diff
@@ -75,20 +75,21 @@ public sealed class ExifReader : IJpegSegmentMetadataReader
...
+                exifTiffHandler.Error("ExifReader TiffProcessingException : " + e.Message);
             }
             catch (IOException e)
             {
                 // TODO what do to with this error state?
-                Debug.WriteLine(e);
+                exifTiffHandler.Error("ExifReader IOException : " + e.Message);
             }
 
             return directories;
```
drewnoakes: Nice solution.

PR #72, opened 2016-11-18
Add Panasonic RAW IFD0 directory

kwhopper:
Attempts to address drewnoakes/metadata-extractor#17 and drewnoakes/metadata-extractor#185 and implements a number of missing Panasonic Makernotes tags.

The most interesting change is it can process the RAW Jpg embedded in the file and outputs that metadata along with the RAW stuff. This and the other changes in ExifTiffHandler are the most significant, so inline some comments/questions if you have any and I'll answer.

### DIFF
--- a/MetadataExtractor.Tests/Formats/Netpbm/NetpbmReaderTest.cs
+++ b/MetadataExtractor.Tests/Formats/Netpbm/NetpbmReaderTest.cs
@@ -52,17 +52,15 @@ namespace MetadataExtractor.Tests.Formats.Netpbm
 
         private static void Verify(string content, int formatType, int width, int height, int? maxVal)
         {
-            using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(content)))
-            {
-                var directory = new NetpbmReader().Extract(stream);
+            var stream = new MemoryStream(Encoding.UTF8.GetBytes(content));
+            var directory = new NetpbmReader().Extract(stream);
 
-                Assert.Equal(formatType, directory.GetInt32(NetpbmHeaderDirectory.TagFormatType));
-                Assert.Equal(width, directory.GetInt32(NetpbmHeaderDirectory.TagWidth));
-                Assert.Equal(height, directory.GetInt32(NetpbmHeaderDirectory.TagHeight));
+            Assert.Equal(formatType, directory.GetInt32(NetpbmHeaderDirectory.TagFormatType));
+            Assert.Equal(width, directory.GetInt32(NetpbmHeaderDirectory.TagWidth));
+            Assert.Equal(height, directory.GetInt32(NetpbmHeaderDirectory.TagHeight));
 
-                if (maxVal != null)
-                    Assert.Equal(maxVal, directory.GetInt32(NetpbmHeaderDirectory.TagMaximumValue));
-            }
+            if (maxVal != null)
+                Assert.Equal(maxVal, directory.GetInt32(NetpbmHeaderDirectory.TagMaximumValue));
         }
     }
 }
--- a/MetadataExtractor.Tests/Formats/Netpbm/NetpbmReaderTest.cs
+++ b/MetadataExtractor.Tests/Formats/Netpbm/NetpbmReaderTest.cs
@@ -52,15 +52,17 @@ namespace MetadataExtractor.Tests.Formats.Netpbm
 
         private static void Verify(string content, int formatType, int width, int height, int? maxVal)
         {
-            var stream = new MemoryStream(Encoding.UTF8.GetBytes(content));
-            var directory = new NetpbmReader().Extract(stream);
+            using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(content)))
+            {
+                var directory = new NetpbmReader().Extract(stream);
 
-            Assert.Equal(formatType, directory.GetInt32(NetpbmHeaderDirectory.TagFormatType));
-            Assert.Equal(width, directory.GetInt32(NetpbmHeaderDirectory.TagWidth));
-            Assert.Equal(height, directory.GetInt32(NetpbmHeaderDirectory.TagHeight));
+                Assert.Equal(formatType, directory.GetInt32(NetpbmHeaderDirectory.TagFormatType));
+                Assert.Equal(width, directory.GetInt32(NetpbmHeaderDirectory.TagWidth));
+                Assert.Equal(height, directory.GetInt32(NetpbmHeaderDirectory.TagHeight));
 
-            if (maxVal != null)
-                Assert.Equal(maxVal, directory.GetInt32(NetpbmHeaderDirectory.TagMaximumValue));
+                if (maxVal != null)
+                    Assert.Equal(maxVal, directory.GetInt32(NetpbmHeaderDirectory.TagMaximumValue));
+            }
         }
     }
 }
--- a/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
+++ b/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
@@ -54,7 +54,7 @@ namespace MetadataExtractor.Formats.Exif
                 case ExifDirectoryBase.TagInteropVersion:
                     return GetInteropVersionDescription();
                 case ExifDirectoryBase.TagOrientation:
-                    return GetOrientationDescription(ExifDirectoryBase.TagOrientation);
+                    return GetOrientationDescription();
                 case ExifDirectoryBase.TagResolutionUnit:
                     return GetResolutionDescription();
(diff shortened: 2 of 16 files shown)


drewnoakes:
@kwhopper you're unstoppable! :) It's Saturday morning here and I'm heading out for the day but should have time to go over this tonight.

MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
```diff
@@ -251,19 +251,6 @@ public string GetYCbCrPositioningDescription()
         }
 
         [CanBeNull]
-        public string GetOrientationDescription()
```
drewnoakes: It's worth keeping this method for consistency with other `Get*Description` methods, though it could delegate to the version on `TagDescriptor`.

MetadataExtractor.Tests/Formats/Netpbm/NetpbmReaderTest.cs
```diff
@@ -52,15 +52,17 @@ public void HeaderParsing()
 
         private static void Verify(string content, int formatType, int width, int height, int? maxVal)
         {
-            var stream = new MemoryStream(Encoding.UTF8.GetBytes(content));
-            var directory = new NetpbmReader().Extract(stream);
+            using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(content)))
```
drewnoakes: There's no need to dispose `MemoryStream`. Can we drop that commit from the PR?

MetadataExtractor/Formats/Exif/PanasonicRawWbInfo2Descriptor.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            ushort wbtype;
+            if (!Directory.TryGetUInt16(tagType, out wbtype))
+                return null;
+
+            return GetLightSourceDescription(wbtype);
+        }
+
+        // EXIF LightSource
+        public string GetLightSourceDescription(ushort wbtype)
```
drewnoakes: Looks like this is in multiple places. The comment suggests it's from the Exif spec, so probably also in a third place. Can we consolidate them in `TagDescriptor` or `ExifBaseDescriptor` as a static helper?

MetadataExtractor/Formats/Exif/makernotes/PanasonicMakernoteDescriptor.cs
```diff
@@ -492,7 +696,7 @@ public string GetConversionLensDescription()
         [CanBeNull]
         public string GetDetectedFacesDescription()
         {
-            return BuildFacesDescription(Directory.GetDetectedFaces());
+            return BuildFacesDescription(Directory.GetDetectedFaces()) ?? "0";
```
drewnoakes: Does this tag always have a value in this descriptor? Elsewhere in the library we return `null` when the tag is not present. This change would suggest no faces were detected if the method were called directly, even though perhaps the tag didn't have a value. Consider changing the implementation of `BuildFacesDescription`, but it should still return `null` if the tag was omitted.

MetadataExtractor/Formats/Exif/makernotes/PanasonicMakernoteDescriptor.cs
```diff
@@ -654,13 +858,13 @@ public string GetWhiteBalanceDescription()
         [CanBeNull]
         public string GetBabyAgeDescription()
         {
-            return Directory.GetAge(PanasonicMakernoteDirectory.TagBabyAge)?.ToFriendlyString();
+            return Directory.GetAge(PanasonicMakernoteDirectory.TagBabyAge)?.ToFriendlyString() ?? "(not set)";
```
drewnoakes: Ditto, re `null`.

MetadataExtractor/Formats/Exif/makernotes/PanasonicMakernoteDescriptor.cs
```diff
@@ -654,13 +858,13 @@ public string GetWhiteBalanceDescription()
...
         }
 
         [CanBeNull]
         public string GetBabyAge1Description()
         {
-            return Directory.GetAge(PanasonicMakernoteDirectory.TagBabyAge1)?.ToFriendlyString();
+            return Directory.GetAge(PanasonicMakernoteDirectory.TagBabyAge1)?.ToFriendlyString() ?? "(not set)";
```
drewnoakes: Ditto, re `null`.

MetadataExtractor/Formats/Tiff/DirectoryTiffHandler.cs
```diff
@@ -46,6 +46,21 @@ protected DirectoryTiffHandler([NotNull] List<Directory> directories, [NotNull]
             Directories.Add(CurrentDirectory);
         }
 
+        protected bool ResetInitialDirectory([NotNull] Directory initialDirectory)
```
drewnoakes: Is there a better way to set the initial directory conditional on the TIFF marker? This feels like a workaround to a problem in the current design.

MetadataExtractor.Tests/Formats/Netpbm/NetpbmReaderTest.cs
kwhopper: Undone now.

MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
kwhopper: It's back, and calls the base version.

MetadataExtractor/Formats/Exif/PanasonicRawWbInfo2Descriptor.cs
kwhopper: Moved to TagDescriptor.

MetadataExtractor/Formats/Exif/makernotes/PanasonicMakernoteDescriptor.cs
kwhopper: all three return null again

MetadataExtractor/Formats/Tiff/DirectoryTiffHandler.cs
```diff
@@ -68,10 +51,14 @@ public void EndingIfd()
...
-            directory.Parent = CurrentDirectory;
-            Directories.Add(directory);
+            // If this is the first directory, don't add to the stack
+            if (CurrentDirectory != null)
+            {
+                _directoryStack.Push(CurrentDirectory);
+                directory.Parent = CurrentDirectory;
+            }
             CurrentDirectory = directory;
+            Directories.Add(CurrentDirectory);
```
kwhopper: PushDirectory does double-duty to help address your previous comment.

kwhopper:
To address your last comments, SetTiffMarker now does the selection of initial directory if one is found based on the marker input. I actually think this simplifies the whole initialization by not requiring the base constructor to expect an initial directory. The only way that this might not work is if SetTiffMarker is not used correctly by implementing classes and/or never called. This seems unlikely for the moment.

MetadataExtractor/Formats/Tiff/DirectoryTiffHandler.cs
drewnoakes: That looks much nicer.

drewnoakes:
Unknown tags are looking like an endangered species!

kwhopper:
Thanks. All of these new tags and directories in .NET only creates a lot of conversion work for Java, so I hope it's ok to keep doing this. However, I figure the more tags this supports the more users you'll get, and then the more problems get found and fixed - which gets more users.  :)

drewnoakes:
Conversion work is fine. Much better to get the coverage in there!

Longer term I'd like to move to a better API. As you well know, adding a new tag involves making changes in several places. Logically I'd like it to be something more like the construction of a single object that contains the metadata for that tag. Those objects could be populated from an external file (XML, or whatever) which could then be shared between Java/.NET implementations.

[I did a little work on this in the new-api branch.](https://github.com/drewnoakes/metadata-extractor-dotnet/commit/74bafc7248ea7ebebaee9ab50852c253bf25165e) It's promising, but a bit messy and might not make sense. I hope the future of the project moves in this direction however.

This new approach would also different types of metadata storage, rather than the current `Dictionary<int,object>`. For example, the new 16-bit binary values ([`ProcessBinary`](https://github.com/drewnoakes/metadata-extractor-dotnet/pull/72/files#diff-106d67665046daf76a4d6027b1249ab5R535)) could be modelled as a byte[] with conversion at read-time.

The new API is also designed with writing metadata in mind.

If you get a chance and have the patience to decode the undocumented pile of code in the commit linked above, it'd be great to get your thoughts.

PR #78, opened 2016-12-13
support Reconyx UltraFire makernote

twest820:
Should be pretty straightforward.  A common base class between the HyperFire and UltraFire makernotes seems desirable but---as tag offsets differ, tag values are of different types and formats, and tag names aren't always the same---is not trivial to express safely and clearly.


drewnoakes:
Looks great, thanks.

PR #80, opened 2016-12-27
sort XMP strings in FileProcessor utility

kwhopper:
This is aimed at the metadata-extractor-images project. It sorts the XMP output strings to make it easier for diff detection with the Java project.

If the code works and if possible, you might update the images metadata txt files after this. The sorted XMP will remove a very large number of diffs with the Java output.

A couple more catch-up things and it might be worth another minor release for both projects. Up to you though.  :-)

### DIFF
--- a/MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
+++ b/MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
@@ -96,17 +96,23 @@ namespace MetadataExtractor.Tools.FileProcessor
                             var xmpDirectory = directory as XmpDirectory;
                             if (xmpDirectory?.XmpMeta != null)
                             {
-                                var wrote = false;
+                                var props = new List<string>();
                                 foreach (var prop in xmpDirectory.XmpMeta.Properties)
                                 {
                                     var value = prop.Value;
                                     if (value?.Length > 512)
                                         value = value.Substring(0, 512) + $" <truncated from {value.Length} characters>";
-                                    writer.WriteLine($"[XMPMeta - {prop.Namespace}] {prop.Path} = {value}");
-                                    wrote = true;
+                                    props.Add($"[XMPMeta - {prop.Namespace}] {prop.Path} = {value}");
                                 }
-                                if (wrote)
+                                if (props.Count > 0)
+                                {
+                                    props.Sort();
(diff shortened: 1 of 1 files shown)


kwhopper:
Here's the background on this. All the specs say the order of xml attributes should be irrelevant. XDocument reads them in document order, but the factories in Java sort the tag attributes. XmpCore reads and outputs in the order that the reader retrieved them, so there is a mismatch between the two libraries.

I thought about this PR a little more, and I wonder if always sorting is the right approach. It might make more sense to add a new arg that turns this on - maybe something like "--sortxmp" or whatever. Some people might only deal in one version of the library and this will change their output needlessly.

I'm happy to add the arg if that's the way you want to go, and with whatever names seems reasonable. I still think this PR needs to happen. Otherwise dotnet/Java output is too difficult to reconcile. Thoughts?

drewnoakes:
I'd been meaning to look into why the ordering differed. This looks good to me. I think it's fine to just order them here in the same way that the Java XMPCore library does. In fact, it might make sense to modify the .NET XmpCore library to match the ordering of the Java library from which it is derived. I don't have a strong feeling either way here.

Would it be enough to just write `xmpDirectory.XmpMeta.Properties.OrderBy(p => p.Namespace).ThenBy(p => p.Path)` in the foreach? That stray `wrote` var looks like it can still go.

kwhopper:
Good point -- this project is trying to mimic Java so maybe extra effort to mimic the sort is the better option. I don't think any external sort will ever work; only attributes appear to be sorted in Java (am not 100% certain on this). But, determining whether a node in XMP was originally an attribute doesn't seem easy or maybe possible.

The first solution I tried was actually in the XmpCore project. After the XDocument is parsed, it looped through all nodes and sorted attributes (using ReplaceAttributes) when a node had more than one. I'll try that code again in the Xmp project and see if attributes are really the only problem. Thanks!

kwhopper:
I ran drewnoakes/xmp-core-dotnet#20 through my local images repo and there were no errors. That doesn't mean it's correct but seems close.

drewnoakes:
Merging the other PR closed this automatically.

Your fix is available in [2.0.0-rc6 on NuGet](https://www.nuget.org/packages/XmpCore/2.0.0-rc6).

PR #83, opened 2017-01-12
Add Follower IFD check for multipage tiff files

kwhopper:
drewnoakes/metadata-extractor#223

- Bug fix in ExifDescriptorBase for a missing subfiletype description.
- Added Page Number (0x0129) tag to ExifDescriptorBase
- First go at detecting and displaying multipage tiff images

The problems I see with this PR:
- The new check in HasFollowerIfd() isn't based on a concrete way of knowing it's an 'image' that I can find so far. It's mostly intuition that the page number tag won't exist in other cases.
- Adding additional ExifIfd0Directory's was my first solution. That seemed wrong since these aren't IFD0 and instead created a new directory (ExifImageDirectory). I can't say either way whether that's a good solution, better name, etc.
- Using "PushDirectory" in this case can create a very long indented output of the structure

It makes significant additions (and corrections actually) to the images repo -- a number of TIFF's were multipage but being loading into Thumbnail directories. Let me know what you think.

### DIFF
--- a/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
+++ b/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
@@ -430,9 +430,10 @@ namespace MetadataExtractor.Formats.Exif
         [CanBeNull]
         public string GetNewSubfileTypeDescription()
         {
-            return GetIndexedDescription(ExifDirectoryBase.TagNewSubfileType, 1,
+            return GetIndexedDescription(ExifDirectoryBase.TagNewSubfileType, 0,
                 "Full-resolution image",
                 "Reduced-resolution image",
+                "Single page of multi-page image",
                 "Single page of multi-page reduced-resolution image",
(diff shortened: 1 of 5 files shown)


drewnoakes:
This looks solid and I'd like to give it a proper look over, but won't have time to do so until this weekend. Thanks for the thorough description too.

Do you think there's a case for introducing a different TiffHandler implementation for TIFF/RAW files, as opposed to Exif?

drewnoakes:
Nice bug fix on the `New Subfile Type` tag! Makes a lot more sense now :)

The PageNumber check looks good. Perhaps the subfile type tag could indicate the directory type, though we don't get around to reading its values until after we have chosen the directory type. That'd require a bit of a redesign. In theory, it should differentiate well between a reduced resolution image and a single page of multi-page image. What you've come up with seems to work well on a wide range of images, and is undoubtedly superior to what was there before.

PR #84, opened 2017-01-15
Describe the page number tag

drewnoakes:
@kwhopper could you review this and, if happy, merge it?

It seems to work on many of the ImageTestSuite TIFF files, however there are quite a few of those where the page numbers make little sense.

I used this page as a reference for the tag: http://www.awaresystems.be/imaging/tiff/tifftags/pagenumber.html

### DIFF
--- a/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
+++ b/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
@@ -58,6 +58,8 @@ namespace MetadataExtractor.Formats.Exif
                     return GetOrientationDescription();
                 case ExifDirectoryBase.TagResolutionUnit:
                     return GetResolutionDescription();
+                case ExifDirectoryBase.TagPageNumber:
+                    return GetPageNumberDescription();
                 case ExifDirectoryBase.TagYCbCrPositioning:
                     return GetYCbCrPositioningDescription();
                 case ExifDirectoryBase.TagXResolution:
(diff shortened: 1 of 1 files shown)


MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
```diff
@@ -271,6 +273,25 @@ public string GetResolutionDescription()
                 "cm");
         }
 
+        [CanBeNull]
+        public string GetPageNumberDescription()
+        {
+            var values = Directory.GetInt32Array(ExifDirectoryBase.TagPageNumber);
```
kwhopper: All the specs I have seen describe this as a short array. Exiftool has it typed as ushort. When feasible, I typically try to keep the same data types in .NET as the spec. Not every array type is implemented, so can fall back on the 'as' operator:

`var values = Directory.GetObject(ExifDirectoryBase.TagPageNumber) as ushort[];`

Obviously, the parsed datatype isn't always to spec so you would have to test whether this works all the time. Doing this in Java isn't always possible, so int32 isn't so bad if you're trying to match code.

kwhopper:
I'm fine with this besides the small datatype quibble. It sounds like these values will either be unknown for the second value or wrong more often than not.

MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
drewnoakes: Agree. We don't have a `GetInt16Array` extension method for `Directory` and I wasn't feeling up to making one, partly because of the inconsistent mess that permitting type coercion within directories has left us with. We could call `GetObject` and cast, but `GetInt32Array` will convert the value if it's non-standard, which does tend to increase coverage.

PR #85, opened 2017-01-21
IPTC output fixes and additions + Deflate fix.

kwhopper:
- port of drewnoakes/metadata-extractor@184ce9c
- refactor IPTCReader.ProcessTag to use StringValue and get closer to Java output
- Added some StringValue methods to DirectoryExtensions and SequentialReader
- DeflateStream implements RFC 1951, but not RFC 1950. Strip first two bytes of input array to make it 'deflate' compatible (should watch for any new issues in the future with this)
- If StringValue's ToInt32 converter can't parse the input as a string, loop through the bytes and make an int out of them (similar to the Java project's getInteger method)

### DIFF
--- a/MetadataExtractor/DirectoryExtensions.cs
+++ b/MetadataExtractor/DirectoryExtensions.cs
@@ -410,6 +410,18 @@ namespace MetadataExtractor
             if (s != null)
                 return new[] { s };
 
+            if (o is StringValue)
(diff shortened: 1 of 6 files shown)


drewnoakes:
Good stuff.

Seems AppVeyor is stuck. Will investigate.

PR #90, opened 2017-02-13
New API exploration

drewnoakes:
This pull request is exploring an improved API within the library the library.

Discussion about the API should happen here.

### DIFF
--- a/MetadataExtractor/ExifTags.cs
+++ b/MetadataExtractor/ExifTags.cs
@@ -1,0 +1,839 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using MetadataExtractor.Util;
+
+namespace MetadataExtractor.NewApi
+{
+    public static class ExifTags
+    {
+        public static readonly TiffStringTag InteropIndex = new TiffStringTag(0x0001, "Interoperability Index", describer: (value, format) =>
+        {
+            return string.Equals("R98", value.Trim(), StringComparison.OrdinalIgnoreCase)
+                ? "Recommended Exif Interoperability Rules (ExifR98)"
+                : value;
+        });
+
+        // TODO the element type here is undefined -- may need coercion
+        public static readonly TiffUInt16ArrayTag InteropVersion = new TiffUInt16ArrayTag(0x0002, "Interoperability Version", 4, (values, format) => DescribeVersion(values, 2));
+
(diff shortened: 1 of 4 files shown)


rcketscientist:
While writing the x3f support which can have string keys I was thinking about this.  While it would require a massive code reformat I  think it would actually have a minimal logical impact to do:

`Key`
|-`StringKey`
|-`IntegerKey`
|-`CompoundKey`

With `Directory<T>`s managing themselves, the type of the constants should be transparent to the user, unless I'm overlooking something.

Sorry, I'm not a coder primarily, so I don't know the fancy terms for what I just tried to describe.

rcketscientist:
(Java)
I scrapped the simple key extension idea.  It worked great for replacing everything with some quick regex, but wasn't compatible with switch statements.  I think all we need is a reverse-lookup enum.

I modified `Directory` and one of the subclasses here:
https://gist.github.com/rcketscientist/60a03908034d653a1c9134f1abaf38e5

I threaded the logic through the descriptor and exif handler without issues.  As you can see `Directory` defers the map logic to its subclasses.  It should handle the change transparently with the addition of reverse-lookup gets:

K is enum.
T is Integer in most cases.
```
    public void setObject(K tagType, @NotNull Object value)
    {
        if (value == null)
            throw new NullPointerException("cannot set a null object");

        if (!getTagMap().containsKey(tagType)) {
            _definedTagList.add(new Tag(tagType, this));
        }
        getTagMap().put(tagType, value);
    }

    public void setObject(T tagValue, @NotNull Object value)
    {
        setObject(getTagFromValue(tagValue), value);
    }
```

**This solution will probably be a little more manual to implement, so let me know what you think before I go refactoring the whole project.**

rcketscientist:
I played with the enum idea a bit more and realized it can tick all the checkboxes in java.  Using interfaces and the template capability in java enums eliminates the need for `Tag`, `Descriptor` and locates all tag information centrally in the enum itself.

Key (interface):
https://gist.github.com/rcketscientist/99830fbbb1ce0e451bbf698aeeca0678

Example directory:
https://gist.github.com/rcketscientist/60a03908034d653a1c9134f1abaf38e5

By consolidating so much we'd lose plug-in backwards compatibility, but updating to the new API shouldn't be too much work.

PR #91, opened 2017-02-20
Implement issue #232

kwhopper:
First try at (most of) drewnoakes/metadata-extractor#232

Consider this ALPHA; I'd really like a lot of feedback before you/we integrate this PR. There is some commented-code I want to keep in there until some review is done.

The main change is obviously no more byte arrays in a jpegsegment. Instead, the absolute location values are stored as @drewnoakes implemented in ShowJpegStructure with a few updates. This implied that two passes would be required, at least in this implementation. The first pass records the segments, the second pass extracts the metadata expecting a reference to the original still-open Stream.

JpegSegmentReader has some notable changes and most are self-explanatory. One relates to drewnoakes/metadata-extractor#234 and processing beyond the first Sos marker. I decided to leave multiple SOS processing in there for this PR, but ultimately I think it should be removed and put the "yield break;" back (it's commented out). ExifTool says that metadata stops after Sos is reached. I tend to agree; DHT data feels outside the scope of this library. But that's certainly open to debate.

Let me know what you guys want to do, and please look over the code if you get some time.


kwhopper:
Update - with buffering, SOS processing is more reasonable and can probably stay. It still adds about 60% extra processing time for the entire images repository, but without buffering it was more like 500%.

Issue #89, opened 2017-02-08
Regression tests should not use machine's locale


drewnoakes:
Developers currently cannot run the regression tests in locales that format things differently to `en-GB` (and some other English-language locales). This was identified by @Nadahar in drewnoakes/metadata-extractor#233.

The regression test program must use invariant culture on its thread:

    System.Threading.Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;

This should be compared with the [equivalent issue](https://github.com/drewnoakes/metadata-extractor/issues/236) in the Java project.

patricksadowski:
I ran into the same issue. May I fix it?

kwhopper:
A PR for review is always welcome. Thanks

PR #96, opened 2017-03-16
Test should depend on hard coded culture.

Tasteful:
This PR is updating a failing test if not using culture that have dot (.) as decimal separator.

Can be related to issue #89


drewnoakes:
Interesting way of solving this. But actually there's an existing approach to dealing with test under different cultures. Thanks for pointing this problem out.

Issue #93, opened 2017-03-15
VS2017 / C# 7 support


drewnoakes:
Last week VS2017 was released along with C# 7.

Does anyone have any reservations about the project upgrading? This would cause the solution to not _build_ under VS2015, however you could still use the NuGet package in earlier version(s) of Visual Studio just fine.

kwhopper:
I'm all for it. VS2017 Community is free anyway if people need to build. Microsoft will likely stop updating VS2015 very soon (if they haven't already) as they did with 2013, and newer .NET Core support might start to dwindle there as well.

That said - you'll have to be careful that C# 7 support doesn't push the .NET Framework version requirement further than you want it to go.

drewnoakes:
Agreed -- we wouldn't change the minimum framework requirement at all.

twest820:
The projects were upgraded to VS2017 yesterday and no longer open in VS2015.3.  However, my VS2017 RTM install is unable to locate any of the tests after a debug or retail build or completes.  The xunit packages installed without issue and VS does see [Fact] as a test marker, enabling Run Tests in the method's right click menu.  But it then doesn't execute any test and produces no test output indicating the cause of the failure.  Known issue?  I checked the wiki but didn't see any instructions about any additional setup requirements and I've other projects where VS2017 runs the tests fine.

With the x86 test engine:
------ Load Playlist started ------
========== Load Playlist finished (0:00:00.0040037) ==========
------ Discover test started ------
========== Discover test finished: 0 found (0:01:01.3259922) ==========

With the x64 test engine:
------ Discover test started ------
Test run will use DLL(s) built for framework Framework45 and platform X64. Following DLL(s) will not be part of run: 
MetadataExtractor.Benchmarks.exe is built for Framework Framework45 and Platform X86.
 Go to http://go.microsoft.com/fwlink/?LinkID=236877&clcid=0x409 for more details on managing these settings.
========== Discover test finished: 0 found (0:01:03.7143222) ==========

drewnoakes:
@twest820 could you try again now that #100 has been merged?

twest820:
[...]

PR #97, opened 2017-03-16
Update to use VS2017 csproj file

Tasteful:
Update to use vs2017 csproj file, see #93


MetadataExtractor.Tests/Formats/Exif/NikonType2MakernoteTest.cs
```diff
@@ -303,11 +303,11 @@ public void GetAutoFlashCompensationDescription()
...
             Assert.Equal("1 EV", descriptor.GetAutoFlashCompensationDescription());
             directory.Set(NikonType2MakernoteDirectory.TagAutoFlashCompensation, new sbyte[] { 0x04, 0x01, 0x06 });
-            Assert.Equal("0.67 EV", descriptor.GetAutoFlashCompensationDescription());
+            Assert.Equal($"{0.67} EV", descriptor.GetAutoFlashCompensationDescription());
             directory.Set(NikonType2MakernoteDirectory.TagAutoFlashCompensation, new sbyte[] { 0x02, 0x01, 0x06 });
-            Assert.Equal("0.33 EV", descriptor.GetAutoFlashCompensationDescription());
+            Assert.Equal($"{0.33} EV", descriptor.GetAutoFlashCompensationDescription());
             directory.Set(NikonType2MakernoteDirectory.TagAutoFlashCompensation, new sbyte[] { unchecked((sbyte)0xFE), 0x01, 0x06 });
-            Assert.Equal("-0.33 EV", descriptor.GetAutoFlashCompensationDescription());
+            Assert.Equal($"{-0.33} EV", descriptor.GetAutoFlashCompensationDescription());
```
drewnoakes: Seems changes from your other PR made it into here.

MetadataExtractor.Tests/Formats/Jpeg/JpegMetadataReaderTest.cs
```diff
@@ -34,7 +34,7 @@ namespace MetadataExtractor.Tests.Formats.Jpeg
     /// <author>Drew Noakes https://drewnoakes.com</author>
     public sealed class JpegMetadataReaderTest
     {
-#if !NETCOREAPP1_0
+#if !NETCOREAPP1_3
```
drewnoakes: Does this have to be raised?

MetadataExtractor.Tests/MetadataExtractor.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp1.3</TargetFramework>
```
drewnoakes: Ditto

MetadataExtractor.Tests/RationalTest.cs
```diff
@@ -109,7 +109,7 @@ public void ZeroOverZero()
             Assert.True(new Rational(0, 0).IsInteger);
         }
 
-#if !NETCOREAPP1_0
+#if !NETCOREAPP1_3
```
drewnoakes: Ditto

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+before_build:
+  appveyor-retry nuget restore
+build:
+  parallel: true
+  verbosity: minimal
+test:
+  assemblies:
+    - MetadataExtractor.Tests.dll
+test_script:
+  - dotnet test .\MetadataExtractor.Tests\MetadataExtractor.Tests.csproj --configuration Release --no-build
```
drewnoakes: Is `test_script` needed here? It wasn't required for [XmpCore](https://github.com/drewnoakes/xmp-core-dotnet/blob/master/appveyor.yml#L12-L14).

MetadataExtractor.Tests/MetadataExtractor.Tests.csproj
Tasteful: The MetadataExtractor is targeting `netstandard1.3` and not `netstandard1.0`, this causes a lot of warning. The test should probably target the same TFM as the library.

appveyor.yml
Tasteful: As standard it was running `xunit.consol`-runner that was working for XmpCore because that the TFS for the test in XmpCore is `netcoreapp1.0;net451` and that runner only execute the classic net451 assembly tests.

In MetadataExtractor the test is only targeting `netcoreapp1.3` and not classic .net and to get the test to run I need to tell the appveyor test-runner to run the `dotnet test` instead.

Before I added the `test_script` property the build was failing, see example https://ci.appveyor.com/project/drewnoakes/metadata-extractor-dotnet/build/2.0.249#L397

MetadataExtractor.Tests/MetadataExtractor.Tests.csproj
drewnoakes: `netcoreapp` and `netstandard` are different things. `netcoreapp1.3` doesn't exist yet, afaik.

appveyor.yml
drewnoakes: Tests could target `net451` as well. Ideally we'd test both build configurations.

Tasteful: I have added so test now also is compiling for net451 but that are not running due to path issue when finding file-assets for metadata extraction.

If we check one of the old builds that was working, example https://ci.appveyor.com/project/drewnoakes/metadata-extractor-dotnet/build/1.0.227 we can see that the test is running netcoreapp1.0-version. 

Make the net451 test running on ci-server can probably be added later and until that the test_script need to be added to run correct test-runner.

MetadataExtractor.Tests/MetadataExtractor.Tests.csproj
Tasteful: Correct, I reverted that change

drewnoakes:
I've squashed your commits and played around with them a bit in #99.

I'll review the changes to get the test project running under net451 separately.

Issue #49, opened 2016-05-12
Sign assemblies


drewnoakes:
From the mailing list:

> For SharePoint deployment I need signed DLLs ('strong named'). NuGet gives me three unsigned dlls.
> 
> With some tools it's possible to create signed DLLs, but the references within the MetadataExtractor.DLL to the dependent DLLs (like XmpCore) seems to stay as references to the unsigned ones.
> 
> Are there strong named DLLs of the MetadataExtractor and the dependent libraries available? Did anyone have a similar problem and solved it somehow?

drewnoakes:
The XmpCore library now supports strong names, so support within MetadataExtractor is not far off.

drewnoakes:
This is now available as:

https://www.nuget.org/packages/MetadataExtractor.StrongName/

PR #98, opened 2017-03-16
Create strong name signed version with conditional compilation

Tasteful:
Contains the changes from #97 and fixing the #49.

Run the `Build.ps1` and artifacts will be placed in `artifacts` folder


drewnoakes:
Pushed as ce4e84b88f9. Thanks :)

drewnoakes:
We'll let this settle for a bit, and get it out as part of 2.0.0 soon.

I'm hoping to review @kwhopper's #91 by then too, but that's a biggie :)

PR #100, opened 2017-03-17
Execute .net45 and .netcoreapp1.0 tests

Tasteful:
When xUnit was releasing 2.2.0 they dropped support for net451, see https://xunit.github.io/release-notes/2017-02-19.html

This PR contains updated TFM to net452 that only is for test-project so the class library will still compuled for lower TFM. You can see that in this https://ci.appveyor.com/project/Tasteful/metadata-extractor-dotnet/build/2.0.3 build.

If removing the `test_script` from `appveyor.yml` only the `net452` build is executing, with the `test_script` both test for `net452` and `netcoreapp1.0` is executed.


drewnoakes:
This is great. Thanks very much. Nice to see the test run in each configuration.

Issue #102, opened 2017-04-03
Only first two images are read from multipage TIFF.


tdhintz:
HasFollowerIfd() is hard-coded to treat multipage TIFF as a single image + thumbnail if the optional ExifDirectoryBase.TagPageNumber tag isn't present.  Thus, a 5 page file only presents tags from the first two.  The page number tag is rarely present.

kwhopper:
Do you know of an image in the image repo or have an image you can share that exhibits this behavior? I'm not saying it isn't an issue, but need one to test. Thanks

tdhintz:
I created an image that can be used for testing.  What is the best way to share it with you?

kwhopper:
If the file is small enough, you should be able to add it through the comment box. "Attach files..." is just below the entry area.

Zip is usually a good idea. A very large file, or a file you want to keep private after sharing, might require a hosting service.

tdhintz:
I made it a pull request.

kwhopper:
@tdhintz - the proposed fix in merged. You can fork/download the main project again and build the new assembly with Visual Studio 2017 if you want to test it. Thanks

drewnoakes:
Closing as fixed.

PR #103, opened 2017-04-03
Sample multipage TIFF.

tdhintz:
For issue #102


drewnoakes:
Thanks very much. We actually have a separate repository for images used in testing, so I've pushed it there.

PR #104, opened 2017-04-03
Always assume follower ifd's point to a thumb or image

kwhopper:
This addresses #102 

It seems like HasFollowerIfd was over-thinking what to do, at least for JPEG and some TIFF formats. The spec indicates 'nextIfdOffset' from TiffReader.cs should point at some kind of image (it's the next 'Image file directory' after all).

This isn't exhaustive, but for now if the current Ifd is IFD0, the next Ifd (IFD1) is a thumbnail. Otherwise, it's treated as another image such as in a multi-page TIFF case. Please give it some thought and let me know.

(I don't think the CanonMakernote comment is valid anymore with all the change to Canon makernote parsing. I could be wrong though)


drewnoakes:
Good stuff!

PR #107, opened 2017-06-21
fix loop issue with Jpeg containing Tiff IFD

munzli:
implicit conversion sometimes ended up in an endless loop that caused cpu to spike

also added problematic image to tests, which can be removed if unwanted


drewnoakes:
Could you add the test image to the image database to in a separate PR? We avoid adding new images to this repro.

Fix looks good though. Nice one.

munzli:
done...

should i remove the test?
also noticed that i accidentally added a using directive, should i also correct that?

drewnoakes:
Sorry for the delay with this. June/July/August were challenging months for my family and I'm still picking up things that slipped through the cracks.

Part of this issue was addressed in e324c9929 which IIRC came via the Java implementation. I'll push a commit including your fix shortly, and possibly include the test image to the other repo. Seems there's another there that triggers this (Sony DSC-P12.jpg).

drewnoakes:
Actually it looks like that image is the one you provided. My mistake. Not sure of the origin story there, but I'll get the rest of this PR merged now for you. Thanks again.

drewnoakes:
This PR had two fixes, one of which was addressed between the PR being submitted and now. I've pushed the other fix as 3e3a3ba5179968d11df6cef5ef65cb3392160dd5 and given you as the author.

PR #113, opened 2017-12-25
Better handling of broken mp4-files

VictorLoktev:
I have a mp4 video file made by camera with no other modifications. The QuickTimeReader crashes with no exif information about the file. It looks like the file ends between atom size and atom type bytes where is not exception handling at all. I put whole code in try-catch and add methods to avoid rising of exceptions for stability and performance.
MetadataExtractor.sln: I have no idea what for  MS VS changes project file on open.

### DIFF
--- a/MetadataExtractor.sln
+++ b/MetadataExtractor.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26430.4
+VisualStudioVersion = 15.0.27130.2010
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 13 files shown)


VictorLoktev:
Support of AVI type added. Code based on https://github.com/drewnoakes/metadata-extractor

drewnoakes:
This looks awesome and I will merge it ASAP! Would be cool if you could use spaces instead of tabs, like the rest of the codebase uses.

VictorLoktev:
Sorry for tabs. I use tabs since 80th. And I actually don't remember why I prefer tabs over spaces :-)
If you didn't merge yet you may replace tabs by space if you want. I need couple of days to reach a computer with the project.

VictorLoktev:
Tabs replaced with spaces and other minor code format changes.

PR #117, opened 2018-01-09
Tabs replaced with spaces and other minor code format changes.

VictorLoktev:


MetadataExtractor/IO/SequentialStreamReader.cs
```diff
@@ -98,5 +98,10 @@ public override bool TrySkip(long n)
                 return false;
             }
         }
+
+        public override bool IsCloserToEnd(long numberOfBytes)
+        {
+            return _stream.Position + numberOfBytes >= _stream.Length;
```
drewnoakes: FYI the only reason I haven't merged this yet is because I want to look into this a bit more. Does `Stream.Length` ever throw or cause the stream to read to the end? Not all streams know their lengths. I'm incredibly busy at the moment but will get to this PR along with others that are pending ASAP.

I'm happy for anyone interested in the project to review PRs. It definitely helps improve quality, and I'm all for it.

VictorLoktev:
I used Stream.Length in public override bool IsCloserToEnd(long numberOfBytes) like someone did in method public override void Skip(long n) just 2 method avove:
`
        public override void Skip(long n)
        {
            if (n < 0)
                throw new ArgumentException("n must be zero or greater.");

            if (_stream.Position + n > **_stream.Length**)
                throw new IOException("Unable to skip past of end of file");

            _stream.Seek(n, SeekOrigin.Current);
        }
`
You are right, not all streams supports Seek or Length. Fortunately all implementations in the project works with files and support these method. If you switch to reading data from internet streams it could bring problems.

drewnoakes:
Thanks for the response. Indeed it looks like if this would be a problem, then it already is. Thanks for a great contribution.

PR #129, opened 2018-03-12
Xmp allow raw data

viceroypenguin:
Two improvements:
1. Adding a .editorconfig, which is now supported by VS 2017. This ensures spacing/tab style consistency from anyone who edits files in this solution going forward.
1. Add `XmpRawData` property. Xmp Core enforces strict compliance with the Xmp standard, but at times, it is necessary to extract xmp data from a file where the provider of the file does not meet the standard. Adding this property allows accessing the data directly so users can access the data if it is non-compliant.
** I am not beholden to using `byte[]` for the raw data. I would be happy to change it to `XDocument` if it would make more sense. This seemed the easiest and most flexible way to provide access to the data.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*.cs]
+indent_size = 4
+indent_style = space
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/MetadataExtractor.sln
+++ b/MetadataExtractor.sln
@@ -5,6 +5,7 @@ VisualStudioVersion = 15.0.27130.2010
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{8CF154EA-6A2C-4BF4-B263-78758F834192}"
 	ProjectSection(SolutionItems) = preProject
+		.editorconfig = .editorconfig
 		.gitattributes = .gitattributes
 		.gitignore = .gitignore
 		appveyor.yml = appveyor.yml
--- a/MetadataExtractor/Formats/Xmp/XmpDirectory.cs
+++ b/MetadataExtractor/Formats/Xmp/XmpDirectory.cs
@@ -25,6 +25,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Xml.Linq;
 using JetBrains.Annotations;
 using XmpCore;
 
@@ -60,6 +61,11 @@ namespace MetadataExtractor.Formats.Xmp
         [CanBeNull]
         public byte[] XmpRawData { get; private set; }
 
+        /// <summary>Root XMP <see cref="XDocument"/> as read from the file.</summary>
+        /// <remarks>Allows processing data from this element if the data is not standards-compliant.</remarks>
+        [CanBeNull]
+        public XDocument Root { get; private set; }
+
         public XmpDirectory()
         {
             SetDescriptor(new XmpDescriptor(this));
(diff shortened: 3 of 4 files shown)


michaelosthege:
In an earlier version of #65 I added an XDocument property too:

In `XmpDirectory`:
```csharp
/// <summary>
/// Root XMP <see cref="XDocument"/> as read from the file.
/// </summary>
[CanBeNull]
public XDocument Root { get; private set; }

public void SetRootDocument([NotNull] XDocument root)
{
    Root = root;
}
```

In `XmpReader`:
```csharp
[NotNull]
public XmpDirectory Extract([NotNull] byte[] xmpBytes, int offset, int length)
{
    var directory = new XmpDirectory();
    try
    {
        // first extract the XML
        var xmpXdoc = XmpMetaFactory.ExtractXDocumentFromBuffer(xmpBytes);
        directory.SetRootDocument(xmpXdoc);

        // now try to parse it
        var xmpMeta = XmpMetaFactory.ParseFromXDocument(directory.Root);
        directory.SetXmpMeta(xmpMeta);
    }
    catch (XmpException e)
    {
        directory.AddError("Error processing XMP data: " + e.Message);
    }
    return directory;
}
```

Maybe you can combine this into your PR ?

viceroypenguin:
Question: Would you like `XDocument` in addition to `XmpRawData` or instead of `XmpRawData`. Personally, I only did the PR because I needed the `XDocument`, so I don't care myself.

viceroypenguin:
Also, I have noticed that the `Extract()` call is made with `offset`s other than `0`, and `length`s other than `xmpBytes.Length`. This means if I pass `xmpBytes` directly to `ExtractXDocumentFromBuffer()`, it will be incorrect in this situations. It also means that my code is incorrect as it is. 

* To add `offset` and `length` would require updating your other library (XmpCore), which I could do if necessary. 
* Alternatively, I can load the bytes actually required into another `byte[]` and then hand it off to `ExtractXDocumentFromBuffer()`.

Please advise.

michaelosthege:
I am interested in the Root too. The question is if one needs to anticipate faulty Xml that can not even be parsed to XDocument.  For this case I would keep both byte[] and XDocument.

Yes, I just noticed this too.
The call should be:
```
ExtractXDocumentFromBuffer(xmpBytes, offset, length);
```

And a PR to xmp-core-dotnet should add a new line behind [XmpMetaFactory:L76](https://github.com/drewnoakes/xmp-core-dotnet/blob/master/XmpCore/XmpMetaFactory.cs#L76)

```csharp
public static XDocument ExtractXDocumentFromBuffer(byte[] buffer, int offset, int length, ParseOptions options = null) => XmpMetaParser.Parse(new ByteBuffer(buffer, offset, length), options);
```

viceroypenguin:
I've updated this PR. Because we need access to the raw bytes here, it made sense to go ahead and do an `Array.Copy()` and hand the new `byte[]` to `ExtractXDocumentFromBuffer()`. Hope this works for you!

michaelosthege:
Hey, @drewnoakes did you have time to look into this PR yet?

drewnoakes:
Hi all, thanks for your patience.

Nit: the `.editorconfig` stuff would ideally be in a separate PR.

My concern here, as with similar approaches to retaining data structures in memory during parsing, is that most users won't need the data, and that keeping it in memory increases overall memory consumption and puts more pressure on the GC by keeping the object alive longer. Large XMP data chunks increase the chance of LOH fragmentation.

I am sympathetic to your requirement here and want to find a solution that addresses the need without sacrificing perf for other users.

An idea that's been discussed before (eg. for thumbnail data) is to remember the offset and length of XMP data within the original stream such that it may be dug out of the original stream in a subsequent pass. For XMP within JPEG, however, it's possible that the XMP data itself is fragmented between multiple locations and must be reassembled prior to parsing.

Given this is a special case scenario, can we devise an API that allows you to specify that this data should be captured ahead of time somehow? This would likely be part of a larger refactor. Something like:

```c#
var extractor = new MetadataExtractor();
extractor.ReplaceHandler(ContainerType.Xmp, new XmpHandler { StoreRawBytes = true });

var directories = extractor.Extract(stream);
```

I'm not sure what it'd mean to have these handlers, nor how we'd model the types of containers/metadata, or whether that'd be a fixed set or an extensible one, or several other questions, but an approach like this feels more general and flexible.

Further I'd like to rework the `Directory` type to make its storage more flexible, and in doing so actually to unpack the `XmlMeta` object, and drop the reference to that too, at least by default.

Thoughts?

I hope to find a decent chunk of time to address the architectural ideas surrounding the future of this library in the coming month or two.

Issue #122, opened 2018-02-07
Allow for streaming operation


j4m3z0r:
Hi there, I'm trying to use MetadataExtractor on an input that is being streamed over the network. I've implemented a System.IO.Stream subclass that reads data from the network and buffers data that has already been read to allow seeking to work.

Unfortunately, MetadataExtractor seems to read in the entire file before processing _any_ of the metadata. It looks like it might be this line that is the culprit:

https://github.com/drewnoakes/metadata-extractor-dotnet/blob/master/MetadataExtractor/Formats/Tiff/TiffReader.cs#L67

I note that there is already a comment there about this exact issue with a TODO on it. Is there some way to fix this without breaking anything? Alternatively, how about adding an option to allow for stream-based processing and just skip that check when it's set?

drewnoakes:
Hey James,

Thanks for the report and the research. You're right, reading the length is a problem. I think this is the cause of #111 too.

The line you highlight is only there to workaround rare and invalid images. It seems wrong to pay this cost for valid images.

However there are several usages of `Length` in [`TiffReader`](https://github.com/drewnoakes/metadata-extractor-dotnet/blob/e324c99296bb3f4ecd058eaa37072242bcfce1c1/MetadataExtractor/Formats/Tiff/TiffReader.cs). All of these are used for validation, and they are hit quite often.

Running the code across the sample image library, we see the following tally for times that reading the `Length` prevented code going wrong.

| Line | Hit Count | Purpose |
|-|-|-|
| 67 | 0 | First IFD offset OOB (attempt fix) |
| 112 | 2 | IFD offset OOB (halt) |
| 132 | 12 | IFD entries would go OOB (halt) |
| 184 | 230 | IFD entry pointer OOB (skip entry) |
| 197 | 0 | IFD entry truncated (skip) |
| 205 | 0 | IFD entry pointer OOB (skip entry) <sup>1</sup> |
| 239 | 39 | 4 bytes after IFD point OOB (halt) |

<sup>1</sup> This is actually a duplicate of L184, which is why it's never hit. I'll look at tidying that up a bit.

The cost of ignoring these valdiation checks could be OOBE, OOME, or just garbage being produced in the output. It's the inability to filter garbage that makes disabling validation a problem to my mind.

These numbers are just for TIFF too (Exif, TIFF, raw files, ...) and there are other usages of `Stream.Length` for other file types too.

[...]

Issue #62, opened 2016-10-07
Excessive bounds checking when reading multi-byte values


drewnoakes:
For example, `ByteArrayReader.GetInt32` validates the index for four bytes. It then calls `GetByte` four times, which causes a further four validations.

One potential solution is to have a non-public `GetByteInternal` method that doesn't validate.

Investigate at the same time whether we can have the `GetByte` calls be inlined. Being virtual probably nixes that.

twest820:
Approaches in the direction of `BitConverter.ToInt32(GetBytes(index, 4))` with `Array.Reverse()` or such for endianness might be worth a look. I haven't profiled this space much as I'm usually working with SIMD intrinsics for such things I but recall finding they're faster than shifts, presumably because it's C# for reinterpret_cast<>. In general, among the various reader classes I've looked at, it appears to me shifting to a more conventional offset + length approach for indexing into a byte[] might be beneficial. In particular, it moves the GetBytes() contract away from having to copy to a new byte[]. However, making much of this performant is contingent on the reader's maintaining it's own byte[] buffer.

Virtual calls can be inlined when the compiler knows it's dealing with a sealed type. Haven't looked to see if this actually happens, but it's a pretty niche optimization regardless.

PR #130, opened 2018-05-02
RandomAccessStream

kwhopper:
This is a massive PR and I'll explain more about it soon. In a nutshell:
- All sequential, byte, and indexed readers have been replaced by a single RandomAccessStream that handles all core reading operations. It uses a capturing-like structure to buffer data as requested.
- ReaderInfo has replaced byte arrays for segment and related operations. This class is insulated from the underlying data through a RandomAccessStream. These objects know their own global position and describe their own length, byte order, and local position purely by index; few to no bytes are buffered out of the underlying RAS. Sequential and Indexed operations are supported.
- Seekable (e.g. FileStream) and Non-seekable (e.g. HttpClient) streams are supported.

Related to:
#36 
#91 
#122 
#125 
#62 
#35


kwhopper:
ToDo: ICC data spread across multiple JPEG segments is not currently implemented. It could be done with a concatenated array but will attempt to do it with ReaderInfo

kwhopper:
Non-seekable example:
```C#
const string url = "https://github.com/drewnoakes/metadata-extractor-images/raw/master/png/mspaint-8x10.png";

static HttpClient client = new HttpClient();

static async Task RunAsync()
{
    using (var response = await client.GetAsync(url, HttpCompletionOption.ResponseHeadersRead))
    using (Stream streamToReadFrom = await response.Content.ReadAsStreamAsync())
    {
        var directories = ImageMetadataReader.ReadMetadata(streamToReadFrom);
        // display directory contents however you want
    }
}

// call it somewhere in your code:
RunAsync().GetAwaiter().GetResult();
```

kwhopper:
Streams of unknown length are no longer a possibility. If the length is unknown and an index is outside the length of the file, the only way to know that is read all the bytes and then throw an exception. It's far more stable to force a length requirement. A length isn't strictly required, but if it isn't passed the Stream must be seekable. The non-seekable example can be implemented for this as follows:
```
const string url = "https://github.com/drewnoakes/metadata-extractor-images/raw/master/png/mspaint-8x10.png";
long streamLength = -1;

static HttpClient client = new HttpClient();

static async Task RunAsync()
{
    using (var response = await client.GetAsync(url, HttpCompletionOption.ResponseHeadersRead))
    using (Stream streamToReadFrom = await response.Content.ReadAsStreamAsync())
    {
        var contentLength = response.Content.Headers?.SingleOrDefault(h => string.Compare("content-length", h.Key, true) == 0)
                                            .Value?.FirstOrDefault();

        if (contentLength != null)
            streamLength = int.Parse(contentLength);

        var directories = ImageMetadataReader.ReadMetadata(streamToReadFrom, streamLength);
        // display directory contents however you want
    }
}

// call it somewhere in your code:
RunAsync().GetAwaiter().GetResult();
```

kwhopper:
The last commit stabilized this implementation, and I believe it's ready for more rigorous review. It still handles the image repository with few notable changes other than even more metadata is accessible now and more errors are handled gracefully.

kwhopper:
Last commit is a port from Java of drewnoakes/metadata-extractor#250

drewnoakes:
@kwhopper I'll be sure to look at this properly asap. It's an impressive chunk of work and I want to understand it properly.

kwhopper:
Ok, great. I'll apologize up front for the number of changes in a single PR. As it got rolling, there was just no avoiding it. Feel free to ask or comment about anything. There's certainly a chance I've missed obvious implementation issues.

drewnoakes:
The changes look good, and I know what kind of a roll one can get into when making changes like this. Let's work out how to move this forward.

I first need to convince myself that the design works in all scenarios we might expect it to. So please give me a bit of time.

Have you read anything about `System.IO.Pipelines`? As with #136 I'm not sure about moving to spans just yet, but in future it could be wonderful for a library like this.

I'd like to give the user-facing API an overhaul to address some of the classes of feedback we receive. It might make sense to do that and release 3.0 before increasing the minimum version requirements and using span in 4.0.

kwhopper:
I have read a bit on Pipelines, but never implemented anything. It does look interesting though. A big motivation for squishing the reader classes down in this PR is how much easier it is to change the hidden machinery. Even if Pipelines proper isn't a good fit for this library, some of the new BCL memory classes it uses could have a benefit.

I agree on the API and am happy to help on it.

kwhopper:
If it helps move this (or something like it) along - would you rather make this PR into a new version of the library? That way the current code can still receive updates. There are a number of older issues that can't be implemented in the current code and many of them are significant enhancements (thumb bytes to disk, a context object with settings, etc). This PR feels like the best chance at the moment for some of those to materialize.

There's also a Java port - drewnoakes/metadata-extractor#361 - which brings both libraries closer to parity. Thanks!

drewnoakes:
@kwhopper I'm sorry for being so slow on this. I will look at it asap. From the description it sounds great. I just want to think through any potential problems before merging.

markegorman:
man I have not a clew what I'm even doing I just try stuff and mess it up I'm very sry I'm new to this

________________________________

kwhopper:
> Initially it seems you were supporting seeking, but removed that support more recently. Can you elaborate on why you decided to do so?

Use of a seekable stream is still supported.  Note the uses of CanSeek in this method:
https://github.com/kwhopper/metadata-extractor-dotnet/blob/c8e990aad2924f35c53d30e855b694faebe952b7/MetadataExtractor/IO/RandomAccessStream.cs#L475

... and this line in particular:

https://github.com/kwhopper/metadata-extractor-dotnet/blob/c8e990aad2924f35c53d30e855b694faebe952b7/MetadataExtractor/IO/RandomAccessStream.cs#L519

What gave you the impression that it had been removed? Maybe I didn't comment something very well. Seek is definitely used to skip over unneeded areas when the stream reports support for it. There are two main goals for the RandomAccessStream class:  1) Do all low-level reading and buffering one place; and 2) Support both seekable and non-seekable streams in one place.

The reason I added a semi-required Length is for non-seekable streams. There have been various attempts in both libraries over the years to parse with an unknown length. None have really had any lasting success; it's a very difficult problem and tends to be all or nothing. My attempts to check for the end without a length was devolving into a mess that wasn't going to work. Ultimately, it would try to keep reading buffers forever and then fail entirely. As you have noted:

>  It is mostly useful for validation. We regularly see invalid data that leads to out-of-bounds offsets. If we're reading, we get OOM exceptions when we try to allocate enormous arrays. If we're seeking, we end up buffering the whole file before working out that anything was wrong. This is another problem without a good workaround.

Note too that the length passed can be fake, such as Int.MaxValue, as it's only real purpose is to give the code some boundary to use. This greatly simplifies the bounds checking for a non-seekable stream and stabilized the code. The trade-off is passing something "reasonable."

No matter what for non-seekable streams, we must buffer either in memory or on disk since something like a NetworkStream is forward-only. However, note that if the index requested is valid the non-seekable stream will only buffer as much as is required and not the entire Length of the stream.

I'm out of town at the moment and will comment more on this later -- unless that explains it. But definitely keep sending questions or comments. I could come up with some ways of testing to verify how buffering works in this PR if that would help. A more direct form of communication would be much better to review this, but we have what we have.

drewnoakes:
Thanks for the fast response. It's the end of the day here, so just a few quick thoughts in response.

> What gave you the impression that it had been removed?

I fired up an HTTP request against an image and passed the stream to `ImageMetadataReader.Read` and it threw this exception:

https://github.com/drewnoakes/metadata-extractor-dotnet/pull/130/files#diff-8eefcf5a9c8bcffb02d354ccb8f87ba9R62

> Note too that the length passed can be fake, such as Int.MaxValue

I hadn't considered that when reading the code. I'll give it another look with that in mind. However it should work out of the box. Pass a stream, get metadata. Users who demand better performance will be motivated to understand what they can do to achieve that. Users who just want metadata will curse and go looking for another library if the most simple API doesn't work for them. Maybe we just use int.MaxValue as the default.

Will take another look soon and keep the dialogue flowing.

kwhopper:
I'll go ahead and change it to int.MaxValue if a length isn't provided and seeking isn't possible. That seems reasonable and is probably the best compromise. As you may have said, there is no perfect solution for a non-seekable stream (that I know of). Weren't there other issues or PR's that tried to deal with this directly? I'm certain there were but can't find them at the moment.

I also wanted to say - RandomAccessStream is solely a buffering layer, and the intent is to have either a single class or interface that does it. I'm not against sub-classing it if there's a need. Right now, the current code seems to work in a single class.

The real value in this is the ReaderInfo class. It tracks the data requests using only index values and leaves RAS to handle all the heavy buffering work. ReaderInfo makes it possible to avoid using so many byte arrays and some redundant array copying; ReaderInfo's can have overlapping buffer index values but will read from the same buffered bytes instead of doubling-up. Since it tracks global position new tasks are possible down the road. RAS was just a means to this end.

kwhopper:
@drewnoakes Editing metadata is a hot topic. I'm convinced the reader structure has to be overhauled before editing is even feasible. I'd like to take another shot at this PR if you have interest in reviewing it.

The RandomAccessStream classes wouldn't have to be used throughout for a good test. I'm thinking it's easier to copy/paste those classes into another PR, make some unit tests to demo what it offers, and leave the rest of the lib alone. Thanks

Issue #134, opened 2018-08-28
Broken metadata segment


pitr555:
Hi, I am using this library to fetch image dimensions from the stream. Recently I've run into an issue with probably broken JPG image. 

There is null value inside the second XMP segment
`<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="Adobe XMP Core 5.5-c002 1.148022, 2012/07/15-18:06:45        "><rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"> ...... 
NULL byte here 
.....
</x:xmpmeta>
`

and ReadMetadata throws "Unsupported Encoding" exception

I there any way to skip those broken segments? I just want to use available dimensions metadata.

kwhopper:
Is there a link to the image so we can check it out? I have an unpublished fix for some specific null byte issues in XML segments but maybe yours is a different variety.

drewnoakes:
@pitr555 can you share an image that reproduces this issue please?

kwhopper:
Closing for now. Reopen if there is still an issue we can address.

PR #135, opened 2018-08-28
Hold xmp array and lazy-load XmpMeta in XmpDirectory

### LINKED ISSUE #109: Unsupported Encoding attempting to read XMP data from Ricoh Theta S image
Hi,

As recommended in https://stackoverflow.com/questions/46610818/unsupported-encoding-error-in-xmp-directory  Here's an attached image that is giving me an Unsupported Encoding error in the Xmp Directory.  I do not actually get a stack trace, just and error in the _errorsList.

The issue is perhaps in XmpCore? I walked through metadata extractor with my debugger and reached a point where it hands off the file to XmpCore to parse and that call returns the error.

Here's a sample image :
r0010025

kwhopper:
Replace null characters (0x00) with a space (0x20) to make XML array acceptable to XDocument/XmlReader in XmpCore.

This commit should fix #109 and #134 

(Note that the IsLegalXmlCharacter function could check across the entire range of valid characters. Decided to start out conservatively)

### DIFF
--- a/MetadataExtractor.Tests/Formats/Xmp/XmpReaderTest.cs
+++ b/MetadataExtractor.Tests/Formats/Xmp/XmpReaderTest.cs
@@ -52,7 +52,7 @@ namespace MetadataExtractor.Tests.Formats.Xmp
         [Fact]
         public void testExtract_PropertyCount()
         {
-            Assert.Equal(179, _directory.XmpValueCount);
+            Assert.Equal(179, _directory.GetInt32(XmpDirectory.TagXmpValueCount));
         }
 
         [Fact]
--- a/MetadataExtractor.Tests/Formats/Xmp/XmpReaderTest.cs
+++ b/MetadataExtractor.Tests/Formats/Xmp/XmpReaderTest.cs
@@ -52,7 +52,7 @@ namespace MetadataExtractor.Tests.Formats.Xmp
         [Fact]
         public void testExtract_PropertyCount()
         {
-            Assert.Equal(179, _directory.GetInt32(XmpDirectory.TagXmpValueCount));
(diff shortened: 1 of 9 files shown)


kwhopper:
I reverted this but kept the changes as a comment.

Instead, this holds the XMP byte array in the XmpDirectory and lazy loads an XmpMeta object on request. If there are failures, this gives the consumer access to the original byte array through TagXmpByteArray

kwhopper:
This was reverted and modified because I saw a number of embedded XMP's that weren't fixed by replacing with spaces. It seems better in the long run to hold the bytes instead and (maybe) provide some sanitation methods for common problems. I don't think we'll ever be able to handle all cases of bad embedded XMP.

drewnoakes:
@kwhopper firstly let me apologise for leaving your PRs so long without any kind of feedback. I started a new role a few weeks back and life before and after that was uncommonly busy. I should have a little more time to devote to the project over the next while.

It's hard to say without looking closely at the bytes and the encodings we're using with them. `\0` is perfectly valid in many encodings, so we can't simply replace them in the byte stream. If there's a single encoding, there are probably some simple rules that can be used for sanitisation. However, most strings of data seen by MetadataExtractor can be expected to have every encoding under the sun...

My concern with retaining arbitrary-length `byte[]`s in memory is that if images contain over 85kB of XMP (not inconceivable) then their `byte[]` must be allocated on the LOH. In workloads that involve processing large numbers of images, the LOH can struggle under such high levels of pressure.

It's been a while since I looked at the project so let me spend some time to get my thoughts together about all the bits we have at the moment and put together a plan.

kwhopper:
This might have potential:
https://www.nuget.org/packages/System.Buffers

PR #136, opened 2018-09-01
Retarget and utilize System.Memory

iamcarbon:
This PR retargets the library to platforms supporting System.Memory and implements an initial set of refactoring to improve performance and bring down memory usage.

This PR also:

- Updates deps to their latest versions
- Updates test to run on .NETSTANDARD2.1
- Ensure that the code pages are registered on .NETSTANDARD2.0
- Makes readers with single Extract method static (eliminating per use allocations)

### DIFF
--- a/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
+++ b/MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
@@ -742,7 +742,7 @@ namespace MetadataExtractor.Formats.Exif
             string[] cfaColors = { "Red", "Green", "Blue", "Cyan", "Magenta", "Yellow", "White" };
 
             var ret = new StringBuilder();
-            ret.Append("[");
+            ret.Append('[');
             for (var pos = 2; pos < end; pos++)
             {
                 if (pattern[pos] <= cfaColors.Length - 1)
@@ -751,11 +751,11 @@ namespace MetadataExtractor.Formats.Exif
                     ret.Append("Unknown");  // indicated pattern position is outside the array bounds
 
                 if ((pos - 2) % pattern[1] == 0)
-                    ret.Append(",");
+                    ret.Append(',');
                 else if (pos != end - 1)
                     ret.Append("][");
             }
-            ret.Append("]");
+            ret.Append(']');
 
             return ret.ToString();
         }
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
@@ -1079,7 +1079,7 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             for (var i = 0; i < values.Length; i++)
             {
                 if (i == 0)
-                    sb.Append((_filters.ContainsKey(values[i]) ? _filters[values[i]] : "[unknown]") + "; ");
+                    sb.Append((_filters.TryGetValue(values[i], out string value) ? value : "[unknown]") + "; ");
                 else if (i == 3)
                     sb.Append("Partial Color " + values[i] + "; ");
                 else if (i == 4)
@@ -1400,7 +1400,7 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             for (var i = 0; i < values.Length; i++)
             {
                 if (i == 0)
-                    sb.Append(_filters.ContainsKey(values[i]) ? _filters[values[i]] : "[unknown]");
+                    sb.Append(_filters.TryGetValue(values[i], out string value) ? value : "[unknown]");
                 else
                     sb.Append(values[i]);
                 sb.Append("; ");
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusRawDevelopment2MakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusRawDevelopment2MakernoteDescriptor.cs
@@ -170,7 +170,7 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             for (var i = 0; i < values.Length; i++)
             {
                 if (i == 0)
-                    sb.Append(_filters.ContainsKey(values[i]) ? _filters[values[i]] : "[unknown]");
+                    sb.Append(_filters.TryGetValue(values[i], out string value) ? value : "[unknown]");
                 else
                     sb.Append(values[i]);
                 sb.Append("; ");
--- a/MetadataExtractor/Formats/Icc/IccDescriptor.cs
+++ b/MetadataExtractor/Formats/Icc/IccDescriptor.cs
@@ -231,7 +231,7 @@ namespace MetadataExtractor.Formats.Icc
                             {
                                 name = Encoding.UTF8.GetString(bytes, ofs, len);
                             }
-                            res.Append(" ").Append(str).Append("(").Append(name).Append(")");
+                            res.Append(' ').Append(str).Append('(').Append(name).Append(')');
                         }
                         return res.ToString();
                     }
--- a/MetadataExtractor/Formats/Icc/IccDescriptor.cs
+++ b/MetadataExtractor/Formats/Icc/IccDescriptor.cs
@@ -91,13 +91,11 @@ namespace MetadataExtractor.Formats.Icc
                 {
                     case IccTagType.Text:
                     {
-#if !NETSTANDARD1_3
                         try
                         {
                             return Encoding.ASCII.GetString(bytes, 8, bytes.Length - 8 - 1);
                         }
                         catch
-#endif
                         {
                             return Encoding.UTF8.GetString(bytes, 8, bytes.Length - 8 - 1);
                         }
--- a/MetadataExtractor/Formats/Png/PngChunkReader.cs
+++ b/MetadataExtractor/Formats/Png/PngChunkReader.cs
@@ -30,14 +30,14 @@ using MetadataExtractor.IO;
 namespace MetadataExtractor.Formats.Png
 {
     /// <author>Drew Noakes https://drewnoakes.com</author>
-    public sealed class PngChunkReader
+    public static class PngChunkReader
     {
         private static readonly byte[] _pngSignatureBytes = { 0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A };
 
         /// <exception cref="PngProcessingException"/>
         /// <exception cref="System.IO.IOException"/>
         [NotNull]
-        public IEnumerable<PngChunk> Extract([NotNull] SequentialReader reader, [CanBeNull] ICollection<PngChunkType> desiredChunkTypes)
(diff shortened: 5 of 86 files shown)


iamcarbon:
@drewnoakes -- This is ready for review.

There's a ton of places we can eliminate allocations and improve performance if this gets accepted.

iamcarbon:
@drewnoakes Checking-in for feedback on this.

drewnoakes:
@iamcarbon firstly thanks for looking into this kind of optimisation. I've been watching the introduction of `Span<T>` and `System.Memory` for a while now and see the power it could bring to a library like this. Exciting times ahead.

My attempts to use spans on other projects so far have all been met with barriers when needing to call APIs that don't accept spans, so you'd end up allocating anyway and copying just to call APIs. This situation is improving, but it will require a massive increase in our minimum requirements.

The current supported TFMs `net35` `net46` `netstandard1.3`.

This PR increases them to `net461` `netstandard2.0`.

The Java library has regularly supported versions of Java from ten years prior. Right now it supports back to Java 6 which was released in 2006. I think Java 1.5 from 2004 works too. A utility like this shouldn't exclude potential users without good cause.

So, is there adequate cause for moving to spans and dropping other frameworks?

The primary argument for making this change is improved performance and reduced allocations. Let's understand how much difference this is making through some benchmark comparisons on a suite of realistic images. I usually just use the entire sample image library, as it's large and varied enough to test systems well.

The move would have to improve performance significantly, and we'd need pretty compelling evidence of that early on in the process.

I'd also like to find some stats on usage of old frameworks.

As I mentioned to @kwhopper, I want to take a little while to re-familiarise myself with all we have right now and determine a path forward.

MetadataExtractor/Formats/Exif/ExifDescriptorBase.cs
```diff
@@ -751,11 +751,11 @@ private static string FormatCFAPattern(int[] pattern)
                     ret.Append("Unknown");  // indicated pattern position is outside the array bounds
 
                 if ((pos - 2) % pattern[1] == 0)
-                    ret.Append(",");
+                    ret.Append(',');
                 else if (pos != end - 1)
                     ret.Append("][");
             }
-            ret.Append("]");
+            ret.Append(']');
```
drewnoakes: The changes of this kind could be pulled out to another PR, which I'd merge with only a quick review.

MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
```diff
@@ -1079,7 +1079,7 @@ public string GetArtFilterEffectDescription()
             for (var i = 0; i < values.Length; i++)
             {
                 if (i == 0)
-                    sb.Append((_filters.ContainsKey(values[i]) ? _filters[values[i]] : "[unknown]") + "; ");
+                    sb.Append((_filters.TryGetValue(values[i], out string value) ? value : "[unknown]") + "; ");
```
drewnoakes: Ditto.

MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
```diff
@@ -1400,7 +1400,7 @@ private string GetFilterDescription(int tagId)
             for (var i = 0; i < values.Length; i++)
             {
                 if (i == 0)
-                    sb.Append(_filters.ContainsKey(values[i]) ? _filters[values[i]] : "[unknown]");
+                    sb.Append(_filters.TryGetValue(values[i], out string value) ? value : "[unknown]");
```
drewnoakes: Ditto.

MetadataExtractor/Formats/Exif/makernotes/OlympusRawDevelopment2MakernoteDescriptor.cs
```diff
@@ -170,7 +170,7 @@ private string GetFilterDescription(int tagId)
             for (var i = 0; i < values.Length; i++)
             {
                 if (i == 0)
-                    sb.Append(_filters.ContainsKey(values[i]) ? _filters[values[i]] : "[unknown]");
+                    sb.Append(_filters.TryGetValue(values[i], out string value) ? value : "[unknown]");
```
drewnoakes: Ditto.

MetadataExtractor/Formats/Icc/IccDescriptor.cs
```diff
@@ -233,7 +231,7 @@ private string GetTagDataString(int tagType)
                             {
                                 name = Encoding.UTF8.GetString(bytes, ofs, len);
                             }
-                            res.Append(" ").Append(str).Append("(").Append(name).Append(")");
+                            res.Append(' ').Append(str).Append('(').Append(name).Append(')');
```
drewnoakes: Ditto.

MetadataExtractor/GeoLocation.cs
```diff
@@ -89,7 +89,7 @@ public static double[] DecimalToDegreesMinutesSeconds(double value)
         /// as a double.
         /// </summary>
         [CanBeNull, Pure]
-        public static double? DegreesMinutesSecondsToDecimal(Rational degs, Rational mins, Rational secs, bool isNegative)
+        public static double? DegreesMinutesSecondsToDecimal(in Rational degs, in  Rational mins, in Rational secs, bool isNegative)
```
drewnoakes: Ditto.

MetadataExtractor/IO/SequentialStreamReader.cs
```diff
@@ -77,7 +77,7 @@ public override void GetBytes(byte[] buffer, int offset, int count)
         public override void Skip(long n)
         {
             if (n < 0)
-                throw new ArgumentException("n must be zero or greater.");
+                throw new ArgumentException("Must be zero or greater.", nameof(n));
```
drewnoakes: Ditto.

MetadataExtractor/Rational.cs
```diff
@@ -39,11 +37,9 @@ namespace MetadataExtractor
     /// Note that any <see cref="Rational"/> with a numerator of zero will be treated as zero, even if the denominator is also zero.
     /// </remarks>
     /// <author>Drew Noakes https://drewnoakes.com</author>
-#if !NETSTANDARD1_3
     [Serializable]
     [TypeConverter(typeof(RationalConverter))]
-#endif
-    public struct Rational : IConvertible, IEquatable<Rational>
+    public readonly struct Rational : IConvertible, IEquatable<Rational>
```
drewnoakes: Ditto (making type `readonly`).

MetadataExtractor/Rational.cs
```diff
@@ -258,12 +254,10 @@ public string ToSimpleString(bool allowDecimal = true, IFormatProvider provider
         /// </remarks>
         /// <param name="other"></param>
         /// <returns></returns>
-        public bool EqualsExact(Rational other) => Denominator == other.Denominator && Numerator == other.Numerator;
+        public bool EqualsExact(in Rational other) => Denominator == other.Denominator && Numerator == other.Numerator;
```
drewnoakes: Ditto.

MetadataExtractor/Rational.cs
```diff
@@ -258,12 +254,10 @@ public string ToSimpleString(bool allowDecimal = true, IFormatProvider provider
...
 
         public override bool Equals(object obj)
         {
-            if (ReferenceEquals(null, obj))
-                return false;
```
drewnoakes: Ditto.

MetadataExtractor/TagDescriptor.cs
```diff
@@ -311,7 +308,7 @@ protected string GetLensSpecificationDescription(int tagId)
             if (values[0] == values[1])
                 sb.Append(values[0].ToSimpleString()).Append("mm");
             else
-                sb.Append(values[0].ToSimpleString()).Append("-").Append(values[1].ToSimpleString()).Append("mm");
+                sb.Append(values[0].ToSimpleString()).Append('-').Append(values[1].ToSimpleString()).Append("mm");
```
drewnoakes: Ditto.

iamcarbon:
If we accept this on a 3.0 branch, the goal would be to improve performance by at least 2x while making a concentrated effort to eliminate allocations over the current release. I don't think it's worth it otherwise.

MetadataExtractor/Formats/Png/PngChunkReader.cs
```diff
@@ -30,14 +30,14 @@
...
-    public sealed class PngChunkReader
+    public static class PngChunkReader
     {
         private static readonly byte[] _pngSignatureBytes = { 0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A };
 
         /// <exception cref="PngProcessingException"/>
         /// <exception cref="System.IO.IOException"/>
         [NotNull]
-        public IEnumerable<PngChunk> Extract([NotNull] SequentialReader reader, [CanBeNull] ICollection<PngChunkType> desiredChunkTypes)
+        public static IEnumerable<PngChunk> Extract([NotNull] SequentialReader reader, [CanBeNull] ICollection<PngChunkType> desiredChunkTypes)
```
kwhopper: Making all these Extract methods static means they will always be loaded at startup - even if the consumer doesn't need it. For example, if I'm not extracting on a PNG file, it will still load this code the first time the library is created.

Making it static would probably help overall efficiency if you're re-using the same library instance against many files in a loop. But I wonder if it's better to do lazy loading instead.

iamcarbon: @kwhopper Are you thinking of static constructors?

kwhopper: Not really. More of a local, initially null static variable in each MetadataReader class that is set the first time it is needed. After that, it would be reused - like a lazy loaded property.

However, that implies each of the MetadataReader classes are either all static already or only created one time from the top. That would have to be checked. Otherwise this might not be effective.

iamcarbon: Static fields and constructors are initialized before their first use. These readers have neither.

I still don't understand what is loaded at runtime.

MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
iamcarbon: Unfortunately, TryGetValue isn't available on .NET3.5. If we drop .NET 3.5, it's a pretty small leap to jump to .NET461.

iamcarbon:
@drewnoakes  -- tried cherry picking some of these improvements but .NET3.5 is holding us back. Going to abandon this for now and revisit again in the future.

Let me know if you decide to bump up the minimum targets in the future.

MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
iamcarbon: No go for NET3.5.

PR #143, opened 2018-10-13
port #367 from Java

kwhopper:
port drewnoakes/metadata-extractor#367 from Java

Note that .NET uses the 'yield return' logic. This needed a null check to make sure it doesn't return null directories. Not an issue in Java because the 'metadata' object is passed around and would never set the directory in this case.


drewnoakes:
Thanks again Kevin.

PR #151, opened 2019-01-08
Added support for rotation of QuickTime meta data.

### LINKED ISSUE #149: rotation missing from quicktime metadata
Hi,
Just like the issue that was patched with the java metadata-extractor library (https://github.com/payton/metadata-extractor/commit/69bb1baacc3faa9c9088b68c3445c65162b8ae2b) I realized through use that rotation is not supported in this library. I assume that the calculation would be similar (ie using the transformation matrix to compute the rotation value in the QuicktimeMetadataReader class)? I have an implementation, but I'm not getting the expected rotation values, so obviously I don't want to create a pull request from it. I'm on the right track though, correct?

deonberlin:
I've added parsing for track rotation for quick time meta data. 

Fixes #149


drewnoakes:
Thank you!

PR #152, opened 2019-01-08
Fixed rotation and matrix values for quicktime.

deonberlin:
Rotation values are now all positive ranging from 0 to 360 degree clockwise.
Fixed reading matrix values.

### DIFF
--- a/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
+++ b/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
@@ -78,17 +78,17 @@ namespace MetadataExtractor.Formats.QuickTime
                 var height = directory.GetInt32(QuickTimeTrackHeaderDirectory.TagHeight);
                 if (width == 0 || height == 0 || directory.GetObject(QuickTimeTrackHeaderDirectory.TagRotation) != null) return;
 
-                var matrix = directory.GetInt32Array(QuickTimeTrackHeaderDirectory.TagMatrix);
-                if (matrix == null || matrix.Length <= 5) return;
+                if (directory.GetObject(QuickTimeTrackHeaderDirectory.TagMatrix) is float[] matrix && matrix.Length > 5)
+                {
+                    var x = matrix[1] + matrix[4];
+                    var y = matrix[0] + matrix[3];
+                    var theta = Math.Atan2(x, y);
+                    var degree = ((180 / Math.PI) * theta) - 45;
+                    if (degree < 0)
+                        degree += 360;
 
-                var x = matrix[1] + matrix[4];
-                var y = matrix[0] + matrix[3];
-                var theta = Math.Atan2(x, y);
-                var degree = ((180 / Math.PI) * theta) - 45;
-                if (degree < 0)
-                    degree += 360;
-
-                directory.Set(QuickTimeTrackHeaderDirectory.TagRotation, degree);
+                    directory.Set(QuickTimeTrackHeaderDirectory.TagRotation, degree);
+                }
             }
 
             void MoovHandler(AtomCallbackArgs a)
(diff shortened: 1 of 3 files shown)


deonberlin:
No idea how to resolve the CodeFactor issue.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -62,7 +62,7 @@ void TrakHandler(AtomCallbackArgs a)
                         directory.Set(QuickTimeTrackHeaderDirectory.TagAlternateGroup, a.Reader.GetUInt16());
                         directory.Set(QuickTimeTrackHeaderDirectory.TagVolume, a.Reader.Get16BitFixedPoint());
                         a.Reader.Skip(2L);
-                        directory.Set(QuickTimeTrackHeaderDirectory.TagMatrix, a.Reader.GetInt32Array(9));
+                        directory.Set(QuickTimeTrackHeaderDirectory.TagMatrix, a.Reader.GetMatrix());
```
drewnoakes: Does this mean `GetInt32Array` is no longer needed? Can you remove it?

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
```diff
@@ -58,5 +58,21 @@ public static decimal Get32BitFixedPoint([NotNull] this SequentialReader reader)
                 reader.GetUInt16(),
                 decimal.Divide(reader.GetUInt16(), ushort.MaxValue));
         }
+
+        public static int[] GetMatrix(this SequentialReader reader)
```
drewnoakes: I think this should return `float[]`, and the below divisions be done with floats rather than ints to avoid losing precision. It might not matter for rotations of 0/90/180/270 degrees, but in general it seems truer to the fixed encoding.

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
```diff
@@ -58,5 +58,21 @@ public static decimal Get32BitFixedPoint([NotNull] this SequentialReader reader)
...
+        {
+            var matrix = new int[9];
+            for (var i = 0; i < matrix.Length; i++)
+            {
+                var val = reader.GetInt32() / ushort.MaxValue;
+                // the right column is fixed 2.30 instead of 16.16
```
drewnoakes: Do you have a reference for this info?

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
deonberlin: You are right, I've changed that by reading the 16.16 signed value more straight forward.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
deonberlin: Removed the function.

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
deonberlin: Added comment with link to the specification.

deonberlin:
I'm not sure about the Java implementation. I've adjusted the .NET rotation calculation, so the values are what to expect and also orientated on the exit-tool implementation.

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
drewnoakes: Perfect, thank you.

drewnoakes:
@deonberlin thanks very much for the update here. Looks great.

PR #164, opened 2019-05-25
Added DJI to ExifTiffHandler.cs, so DJI makernotes are parsed

AkosLukacs:
First: Thanks for the work you put into this library!

DJI Makernote parsing was already in the code base, but wasn't called from ExifTiffHandler.cs. If I understand the code correctly :)
Added the model check & tested with photos from DJI FC6520, FC6310 and FC330, and works great.

And an unrelated questions: looks like there is activity in this repo, but there hasn't been a release since 2.0.0. Am I missing something? Anything I could help?


drewnoakes:
Thanks very much!

> there hasn't been a release since 2.0.0.

I have been utterly swamped for the last few months with a new job and moving country. The move is done now and I am starting to engage with the project again. I'll run some tests and get a release out of both this .NET version and the Java version soon<sup>TM</sup>.

PR #179, opened 2019-07-16
Fix Avi Duration output hour. "HH" is 24 hour time

kwhopper:
used Ticks because TimeSpan().ToString(string format) isn't available in netstandard1.3

### DIFF
--- a/MetadataExtractor/Formats/Avi/AviRiffHandler.cs
+++ b/MetadataExtractor/Formats/Avi/AviRiffHandler.cs
@@ -112,7 +112,7 @@ namespace MetadataExtractor.Formats.Avi
                             int hours = (int)duration / (int)(Math.Pow(60, 2));
                             int minutes = ((int)duration / (int)(Math.Pow(60, 1))) - (hours * 60);
                             int seconds = (int)Math.Round((duration / (Math.Pow(60, 0))) - (minutes * 60));
-                            string time = new DateTime(2000, 1, 1, hours, minutes, seconds).ToString("hh:mm:ss");
+                            string time = new DateTime(new TimeSpan(hours, minutes, seconds).Ticks).ToString("HH:mm:ss");
 
                             directory.Set(AviDirectory.TAG_DURATION, time);
                             directory.Set(AviDirectory.TAG_VIDEO_CODEC, fccHandler);


MetadataExtractor/Formats/Avi/AviRiffHandler.cs
```diff
@@ -112,7 +112,7 @@ public void ProcessChunk(string fourCc, byte[] payload)
                             int hours = (int)duration / (int)(Math.Pow(60, 2));
                             int minutes = ((int)duration / (int)(Math.Pow(60, 1))) - (hours * 60);
                             int seconds = (int)Math.Round((duration / (Math.Pow(60, 0))) - (minutes * 60));
-                            string time = new DateTime(2000, 1, 1, hours, minutes, seconds).ToString("hh:mm:ss");
+                            string time = new DateTime(new TimeSpan(hours, minutes, seconds).Ticks).ToString("HH:mm:ss");
```
drewnoakes: There's no need to use `DateTime` here at all:

```suggestion
                            string time = new TimeSpan(hours, minutes, seconds).ToString("hh\\:mm\\:ss");
```

https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-timespan-format-strings

Not sure why we have to escape `:` for `TimeSpan` but not `DateTime`.

kwhopper: Are you sure this works for .NET 3.5? This solution build error (in VS2017) is why I still used Ticks and DateTime:

> Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1501	No overload for method 'ToString' takes 1 arguments	MetadataExtractor(net35)	[redacted path]\metadata-extractor-dotnet\MetadataExtractor\Formats\Avi\AviRiffHandler.cs	115	Active

drewnoakes: No I'm not sure at all :)

Looking at the docs you're absolutely right. No `TimeSpan.ToString` overload accepts arguments in `net35`. Sorry for the noise.

drewnoakes: (I see I missed this detail in your description too 🤦‍♂)

PR #180, opened 2019-07-16
Fix NikonType2 GetPowerUpTimeDescription. It isn't an int64

kwhopper:
The Java version of this should get updated with the null check

### DIFF
--- a/MetadataExtractor/Formats/Exif/makernotes/NikonType2MakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/NikonType2MakernoteDescriptor.cs
@@ -22,7 +22,9 @@
 //
 #endregion
 
+using System;
 using System.Diagnostics.CodeAnalysis;
+using MetadataExtractor.Util;
 using JetBrains.Annotations;
 
 namespace MetadataExtractor.Formats.Exif.Makernotes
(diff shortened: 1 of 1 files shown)


MetadataExtractor/Formats/Exif/makernotes/NikonType2MakernoteDescriptor.cs
```diff
@@ -98,7 +100,17 @@ public override string GetDescription(int tagType)
...
+            // the format is : first 2 bytes together are the year, and then each byte after
+            //                 is month, day, hour, minute, second with the eighth byte unused
+            // e.g., 2011:04:25 01:54:58
+
+            var o = Directory.GetByteArray(NikonType2MakernoteDirectory.TagPowerUpTime);
+            if (o == null)
+                return null;
+
+            ushort year = ByteConvert.FromBigEndianToNative(BitConverter.ToUInt16(o, 0));
+            return string.Format($"{year}:{o[2]:D2}:{o[3]:D2} {o[4]:D2}:{o[5]:D2}:{o[6]:D2}");
```
drewnoakes: I was thinking this is a pretty weird date format, but it's consistent with other dates the library returns via descriptors (it comes from the fact that Exif was the first format implemented). Do you think it'd make sense for descriptor methods to return culture-specific date strings? If so I'll file a follow-up issue.

drewnoakes:
> The Java version of this should get updated with the null check

Would you please open an issue against that project so we don't forget?

PR #181, opened 2019-07-16
Port GifControlDirectory additions from Java

kwhopper:
Todo: backport GifControlDescriptor "{value}" addition to Java
Todo: backport to Java GifReader --  directory.Set(GifControlDirectory.TagUserInputFlag, (packedFields & 2) == 2);

### DIFF
--- a/MetadataExtractor/Formats/Gif/GifReader.cs
+++ b/MetadataExtractor/Formats/Gif/GifReader.cs
@@ -329,12 +329,12 @@ namespace MetadataExtractor.Formats.Gif
 
             var directory = new GifControlDirectory();
 
-            reader.Skip(1);
-
+            byte packedFields = reader.GetByte();
+            directory.Set(GifControlDirectory.TagDisposalMethod, (packedFields >> 2) & 7);
+            directory.Set(GifControlDirectory.TagUserInputFlag, (packedFields & 2) == 2);
+            directory.Set(GifControlDirectory.TagTransparentColorFlag, (packedFields & 1) == 1);
             directory.Set(GifControlDirectory.TagDelay, reader.GetUInt16());
-
-            if (blockSizeBytes > 3)
-                reader.Skip(blockSizeBytes - 3);
+            directory.Set(GifControlDirectory.TagTransparentColorIndex, reader.GetByte());
 
             // skip 0x0 block terminator
             reader.GetByte();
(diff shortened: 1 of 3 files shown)


MetadataExtractor/Formats/Gif/GifReader.cs
```diff
@@ -329,12 +329,12 @@ private static GifControlDirectory ReadControlBlock(SequentialReader reader, byt
...
-
+            byte packedFields = reader.GetByte();
+            directory.Set(GifControlDirectory.TagDisposalMethod, (packedFields >> 2) & 7);
+            directory.Set(GifControlDirectory.TagUserInputFlag, (packedFields & 2) == 2);
+            directory.Set(GifControlDirectory.TagTransparentColorFlag, (packedFields & 1) == 1);
             directory.Set(GifControlDirectory.TagDelay, reader.GetUInt16());
-
-            if (blockSizeBytes > 3)
-                reader.Skip(blockSizeBytes - 3);
+            directory.Set(GifControlDirectory.TagTransparentColorIndex, reader.GetByte());
```
drewnoakes: This used to sometimes skip three bytes, but now always consumes only one. Is this correct?

kwhopper: I don't know for sure without further investigation. This was ported directly from the Java version approved through drewnoakes/metadata-extractor#241.

Maybe @Nadahar has some info on it?

Nadahar: I'm sorry, I checked in my memory, but I'm unable to revive any information about this. From what I can see from the PR, additional parsing was added. It doesn't seem unreasonable that less will be skipped when more is parsed.. does this behavior cause a problem?

drewnoakes: > additional parsing was added

It's just unpacking a single byte that was previously skipped.

@kwhopper if you're happy that your testing produced better results with this change then I'm happy.

kwhopper: It appears to produce the same results as the Java implementation against the image repository. To me, that's the best we can reasonably export when porting.

drewnoakes: Thank you for confirming. Merged!

drewnoakes:
> Todo: backport GifControlDescriptor "{value}" addition to Java
> Todo: backport to Java GifReader -- directory.Set(GifControlDirectory.TagUserInputFlag, (packedFields & 2) == 2);

Could you open issue for these two please so they're not lost?

PR #182, opened 2019-07-16
Port of #325 from Java

kwhopper:
Port of drewnoakes/metadata-extractor#325 :

"Problem:
PngDescriptor.GetBackgroundColorDescription() always returns null even if "TagBackgroundColor" has some value. This is because the code checks the "TagColorType" value, but it exists in the different directory (PNG-IHDR).

Solution:
Return the description based on the length of the byte array."


drewnoakes:
Would you be able to port the corresponding unit tests too?

https://github.com/drewnoakes/metadata-extractor/pull/325/files#diff-57f370cfb4dbb3759bf55d89f8982333

kwhopper:
Done, and pushed

PR #183, opened 2019-07-16
catch PngMetadataReader iCCP decompression errors to match Java output

kwhopper:
Todo on Java - investigate whether this line is redundant:

https://github.com/drewnoakes/metadata-extractor/blob/81143f746e6774eb7625e374b3f6e56396288d1a/Source/com/drew/imaging/png/PngMetadataReader.java#L180


drewnoakes:
> Todo on Java - investigate whether this line is redundant:

It looks like it is redundant. Can you file a follow up issue or PR?

It might make sense for `Metadata.addDirectory` to just return if a given `Directory` instance is already present. Conceptually it's a set of directories.

PR #184, opened 2019-07-16
use more appropriate Skip

kwhopper:


kwhopper:
Todo: The Java implementation of the RiffReader.processChunks has almost no error condition checks for things like too short of a header (8 bytes) or tracking sizeLeft. Should be reviewed:

https://github.com/drewnoakes/metadata-extractor/blob/55651c2918eff6c53485b2216da86e6e6bf2c18a/Source/com/drew/imaging/riff/RiffReader.java#L78

drewnoakes:
> Todo: The Java implementation...

Please file an issue against that project so the work is not forgotten.

PR #185, opened 2019-07-16
Port extended BMP support from Java

kwhopper:
I'm not really impressed with how the enums + description worked in the Java implementation, but it is what it is. Maybe that should be reviewed a little. Using the special enum methods that Java supports kind of breaks the mold for how the library stores and returns descriptions.


drewnoakes:
Regarding the nested enums, they could be nested static classes with constant fields. That'd avoid the need for casts. I don't think the enum types here are actually used anywhere, other than to dereference their members. IIRC we do that for one of the makernotes (Canon I think).

PR #188, opened 2019-07-22
Better handling of collections in descriptors

drewnoakes:


kwhopper:
Is there some test case that is fixed/improved by this? Just curious, and wondered what prompted the change.

drewnoakes:
True, I should add a test. I'll do that, thanks.

I saw something like `System.Generic.Collections.List`1@12345` in the output of a .heic file (IIRC).

PR #189, opened 2019-07-22
Java port - Process only leafnodes for XMP data

kwhopper:
- process only leafnodes for XMP data both in XmpDirectory and text file output
- use consistent new lines in TextFileOuputHandler

Before this change, .NET would go directly through XmpDirectory.XmpMeta for normal processing and display. Namespace nodes are also present in this collection and output lines with blank values. This isn't an error, but they're always like this and Java ignores them by setting an iterator option. With the change, text file output is consistent with Java and removes scores of discrepancies.

Note that XmpMeta is using an Iterator instance with a null IteratorOptions parameter internally for the Properties property, meaning there is no more of a performance hit for this than calling Properties.

### DIFF
--- a/MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
+++ b/MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
@@ -111,18 +111,15 @@ namespace MetadataExtractor.Tools.FileProcessor
                                 while (iterator.HasNext())
                                 {
                                     var prop = (IXmpPropertyInfo)iterator.Next();
-                                    var ns = prop.Namespace;
                                     var path = prop.Path;
-                                    var value = prop.Value;
 
                                     if (path == null)
                                         continue;
-                                    if (ns == null)
-                                        ns = "";
 
-                                    if (value == null)
-                                        value = "";
-                                    else if (value.Length > 512)
+                                    var ns = prop.Namespace ?? "";
+                                    var value = prop.Value ?? "";
+
+                                    if (value.Length > 512)
                                         value = value.Substring(0, 512) + $" <truncated from {value.Length} characters>";
                                     writer.Write($"[XMPMeta - {ns}] {path} = {value}{NEW_LINE}");
                                     wrote = true;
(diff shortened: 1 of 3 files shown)


MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
```diff
@@ -87,27 +93,42 @@ public override void OnExtractionSuccess(string filePath, IList<Directory> direc
...
                                     var value = prop.Value;
-                                    if (value?.Length > 512)
+
+                                    if (path == null)
+                                        continue;
+                                    if (ns == null)
+                                        ns = "";
+
+                                    if (value == null)
+                                        value = "";
```
drewnoakes: You could null-coalesce above instead:

```c#
var ns = prop.Namespace ?? "";
var value = prop.Value ?? "";
```

MetadataExtractor/Formats/Xmp/XmpDirectory.cs
```diff
@@ -76,17 +78,43 @@ protected override bool TryGetTagName(int tagType, out string tagName)
...
-            return XmpMeta?.Properties
-                       .Where(p => p.Path != null)
-                       .ToDictionary(p => p.Path, p => p.Value)
-                   ?? new Dictionary<string, string>();
+            var propertyValueByPath = new Dictionary<string, string>();
+            if (XmpMeta != null)
+            {
+                try
+                {
+                    XmpIterator i = new XmpIterator((XmpMeta)XmpMeta, null, null, new IteratorOptions() { IsJustLeafNodes = true });
```
drewnoakes: (Wow, `XmpIterator` has an ugly API!)

You could cache an instance of `IteratorOptions` somewhere static if you wanted, to reduce allocations.

MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
kwhopper: Good call. Done

MetadataExtractor/Formats/Xmp/XmpDirectory.cs
kwhopper: Also done

drewnoakes:
The unit tests are failing. Please verify that the actual values are correct then update the tests.

kwhopper:
Unit tests are fixed. The expected value in dotnet (167) is different from Java (168) because XmpCore 6.1.10 does some aliasing internally and ignores one of the keys.

PR #191, opened 2019-07-23
Small bits

drewnoakes:


MetadataExtractor.Tools.FileProcessor/TextFileOutputHandler.cs
```diff
@@ -42,7 +42,7 @@ internal class TextFileOutputHandler : FileHandlerBase
         public override void OnStartingDirectory(string directoryPath)
         {
             base.OnStartingDirectory(directoryPath);
-            System.IO.Directory.Delete(Path.Combine(directoryPath, "metadata"), recursive: true);
+            System.IO.Directory.Delete(Path.Combine(directoryPath, "metadata\\dotnet"), recursive: true);
```
kwhopper: This could be more readable with a verbatim string literal:
```suggestion
            System.IO.Directory.Delete(Path.Combine(directoryPath, @"metadata\dotnet"), recursive: true);
```

PR #197, opened 2019-08-19
Port of Java #231: Implement Huffman Tables directory and reader

kwhopper:
Port of drewnoakes/metadata-extractor#231

### DIFF
--- a/MetadataExtractor/Formats/Jpeg/HuffmanTablesDescriptor.cs
+++ b/MetadataExtractor/Formats/Jpeg/HuffmanTablesDescriptor.cs
@@ -1,7 +1,6 @@
 #region License
 //
 // Copyright 2002-2019 Drew Noakes
-// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
--- a/MetadataExtractor/Formats/Jpeg/HuffmanTablesDescriptor.cs
+++ b/MetadataExtractor/Formats/Jpeg/HuffmanTablesDescriptor.cs
@@ -1,0 +1,69 @@
+#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+// More information about this project is available at:
+//
+//    https://github.com/drewnoakes/metadata-extractor-dotnet
+//    https://drewnoakes.com/code/exif/
+//
+#endregion
+
+using System.Diagnostics.CodeAnalysis;
+using JetBrains.Annotations;
+
+namespace MetadataExtractor.Formats.Jpeg
+{
+    /// <summary>Provides a human-readable string version of the tag stored in a HuffmanTablesDirectory.</summary>
+    /// <remarks>
+    /// Provides a human-readable string versions of the tag stored in a HuffmanTablesDirectory.
+    /// <list type="bullet">
+    ///   <item>https://en.wikipedia.org/wiki/Huffman_coding</item>
+    ///   <item>http://stackoverflow.com/a/4954117</item>
+    /// </list>
+    /// </remarks>
+    /// <author>Nadahar</author>
+    /// <author>Kevin Mott https://github.com/kwhopper</author>
+    [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
+    public sealed class HuffmanTablesDescriptor : TagDescriptor<HuffmanTablesDirectory>
+    {
+        public HuffmanTablesDescriptor([NotNull] HuffmanTablesDirectory directory)
+            : base(directory)
+        {
+        }
+
+        public override string GetDescription(int tagType)
+        {
+            switch (tagType)
+            {
+                case HuffmanTablesDirectory.TagNumberOfTables:
+                    return GetNumberOfTablesDescription();
+                default:
+                    return base.GetDescription(tagType);
+            }
+        }
+
+        [CanBeNull]
+        public string GetNumberOfTablesDescription()
+        {
+            if (!Directory.TryGetInt32(HuffmanTablesDirectory.TagNumberOfTables, out int value))
+                return null;
+
+            return value + (value == 1 ? " Huffman table" : " Huffman tables");
+        }
+
+    }
+}
--- a/MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
+++ b/MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
@@ -1,7 +1,6 @@
 #region License
 //
 // Copyright 2002-2019 Drew Noakes
-// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -40,32 +39,32 @@ namespace MetadataExtractor.Formats.Jpeg
     {
         public const int TagNumberOfTables = 1;
 
-        internal static byte[] TypicalLuminanceDcLengths = {
+        public static readonly byte[] TypicalLuminanceDcLengths = {
             0x00, 0x01, 0x05, 0x01, 0x01, 0x01, 0x01, 0x01,
             0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
         };
 
-        internal static byte[] TypicalLuminanceDcValues = {
+        public static readonly byte[] TypicalLuminanceDcValues = {
             0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
             0x08, 0x09, 0x0A, 0x0B
         };
 
-        internal static byte[] TypicalChrominanceDcLengths = {
+        public static readonly byte[] TypicalChrominanceDcLengths = {
             0x00, 0x03, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
             0x01, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00
         };
 
-        internal static byte[] TypicalChrominanceDcValues = {
+        public static readonly byte[] TypicalChrominanceDcValues = {
             0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
             0x08, 0x09, 0x0A, 0x0B
         };
 
-        internal static byte[] TypicalLuminanceAcLengths = {
+        public static readonly byte[] TypicalLuminanceAcLengths = {
             0x00, 0x02, 0x01, 0x03, 0x03, 0x02, 0x04, 0x03,
             0x05, 0x05, 0x04, 0x04, 0x00, 0x00, 0x01, 0x7D
         };
 
-        internal static byte[] TypicalLuminanceAcValues = {
+        public static readonly byte[] TypicalLuminanceAcValues = {
             0x01, 0x02, 0x03, 0x00, 0x04, 0x11, 0x05, 0x12,
             0x21, 0x31, 0x41, 0x06, 0x13, 0x51, 0x61, 0x07,
             0x22, 0x71, 0x14, 0x32, 0x81, 0x91, 0xA1, 0x08,
@@ -89,12 +88,12 @@ namespace MetadataExtractor.Formats.Jpeg
             0xF9, 0xFA
         };
 
-        internal static byte[] TypicalChrominanceAcLengths = {
+        public static readonly byte[] TypicalChrominanceAcLengths = {
             0x00, 0x02, 0x01, 0x02, 0x04, 0x04, 0x03, 0x04,
             0x07, 0x05, 0x04, 0x04, 0x00, 0x01, 0x02, 0x77
         };
 
-        internal static byte[] TypicalChrominanceAcValues = {
+        public static readonly byte[] TypicalChrominanceAcValues = {
             0x00, 0x01, 0x02, 0x03, 0x11, 0x04, 0x05, 0x21,
             0x31, 0x06, 0x12, 0x41, 0x51, 0x07, 0x61, 0x71,
             0x13, 0x22, 0x32, 0x81, 0x08, 0x14, 0x42, 0x91,
@@ -141,7 +140,7 @@ namespace MetadataExtractor.Formats.Jpeg
         [NotNull]
         public HuffmanTable GetTable(int tableNumber)
         {
-            return Tables[tableNumber];
+            return _tables[tableNumber];
         }
 
         /// <returns>The number of Huffman tables held by this HuffmanTablesDirectory instance.</returns>
@@ -150,16 +149,16 @@ namespace MetadataExtractor.Formats.Jpeg
             return this.GetInt32(TagNumberOfTables);
         }
 
-        internal void AddTable(HuffmanTable table)
+        public void AddTable(HuffmanTable table)
         {
-            Tables.Add(table);
+            _tables.Add(table);
             // update the number-of-tables tag with the current count
-            Set(TagNumberOfTables, Tables.Count);
+            Set(TagNumberOfTables, _tables.Count);
         }
 
         /// <returns>The List of HuffmanTables in this Directory.</returns>
         [NotNull]
-        private List<HuffmanTable> Tables { get; set; } = new List<HuffmanTable>(4);
+        private List<HuffmanTable> _tables = new List<HuffmanTable>(4);
 
         /// <summary>Evaluates whether all the tables in this HuffmanTablesDirectory are "typical" Huffman tables.</summary>
         /// <remarks>
@@ -182,11 +181,11 @@ namespace MetadataExtractor.Formats.Jpeg
         /// </remarks>
         public bool IsTypical()
         {
-            if (Tables.Count == 0)
+            if (_tables.Count == 0)
             {
                 return false;
             }
-            foreach (HuffmanTable table in Tables)
+            foreach (HuffmanTable table in _tables)
             {
                 if (!table.IsTypical())
                 {
--- a/MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
+++ b/MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
@@ -25,7 +25,10 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using JetBrains.Annotations;
+
+using MetadataExtractor.Util;
 
 namespace MetadataExtractor.Formats.Jpeg
 {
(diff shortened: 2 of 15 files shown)


MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,323 @@
...
+
+        /// <returns>The number of Huffman tables held by this HuffmanTablesDirectory instance.</returns>
+        public int GetNumberOfTables()
+        {
+            return this.GetInt32(TagNumberOfTables);
+        }
+
+        /// <returns>The List of HuffmanTables in this Directory.</returns>
+        [NotNull]
+        public List<HuffmanTable> Tables { get; set; } = new List<HuffmanTable>(4);
```
drewnoakes: I wouldn't make a mutable `List<>` public here. Is this needed, given you've added `GetTable` and `GetNumberOfTables`? If so, you could make the list a field and change the property to return `IReadOnlyList<HuffmanTable>`.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// </returns>
+        public bool IsOptimized()
+        {
+            return !IsTypical();
+        }
+
+    }
+
+    /// <summary>An instance of this class holds a JPEG Huffman table.</summary>
+    public class HuffmanTable
```
drewnoakes:
```suggestion
    public readonly struct HuffmanTable
```

We'll need to remove the private setters on the fields too.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,323 @@
...
+    {
+        private byte[] _lengthBytes;
+        private byte[] _valueBytes;
+
+        public HuffmanTable([NotNull] HuffmanTableClass tableClass, int tableDestinationId, [NotNull] byte[] lengthBytes, [NotNull] byte[] valueBytes)
+        {
+            _lengthBytes = lengthBytes ?? throw new ArgumentNullException("lengthBytes cannot be null.");
```
drewnoakes: The argument here (and the line below) are supposed to be the argument name. The exception formulates an appropriate string in the user's language.

```suggestion
            _lengthBytes = lengthBytes ?? throw new ArgumentNullException(nameof(lengthBytes));
```

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,323 @@
...
+
+        /// <returns>A byte array with the L values for this table.</returns>
+        [NotNull]
+        public byte[] LengthBytes
+        {
+            get
+            {
+                byte[] result = new byte[_lengthBytes.Length];
+                Array.Copy(_lengthBytes, 0, result, 0, _lengthBytes.Length);
+                return result;
```
drewnoakes: Can this just be `_lengthBytes.ToArray()`?

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,323 @@
...
+                    Equals(_lengthBytes, HuffmanTablesDirectory.TypicalChrominanceDcLengths) &&
+                    Equals(_valueBytes, HuffmanTablesDirectory.TypicalChrominanceDcValues);
+            }
+            else if (TableClass == HuffmanTableClass.AC)
+            {
+                return
+                    Equals(_lengthBytes, HuffmanTablesDirectory.TypicalLuminanceAcLengths) &&
+                    Equals(_valueBytes, HuffmanTablesDirectory.TypicalLuminanceAcValues) ||
+                    Equals(_lengthBytes, HuffmanTablesDirectory.TypicalChrominanceAcLengths) &&
+                    Equals(_valueBytes, HuffmanTablesDirectory.TypicalChrominanceAcValues);
```
drewnoakes: I don't believe these eight `Equals` checks are actually checking the contents of the arrays within.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
```diff
@@ -0,0 +1,106 @@
...
+            directory.Set(HuffmanTablesDirectory.TagNumberOfTables, directory.Tables.Count);
+        }
+
+        private byte[] GetBytes([NotNull] SequentialReader reader, int count)
+        {
+            byte[] bytes = new byte[count];
+            for (int i = 0; i < count; i++)
+            {
+                byte b = reader.GetByte();
+                if ((b & 0xFF) == 0xFF)
```
drewnoakes: I think this can this just be `b == 0xFF`.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    byte stuffing = reader.GetByte();
+                    if (stuffing != 0x00)
+                    {
+                        throw new IOException("Marker " + (JpegSegmentType)stuffing + " found inside DHT segment");
```
drewnoakes: This is not an `IOException`. Maybe `FormatException` or `MetadataException`.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// <seealso cref="JpegSegment"/>
+    /// <author>Nadahar</author>
+    /// <author>Kevin Mott https://github.com/kwhopper</author>
+    public sealed class JpegDnlReader : IJpegSegmentMetadataReader
+    {
+        ICollection<JpegSegmentType> IJpegSegmentMetadataReader.SegmentTypes => new[] { JpegSegmentType.Dnl };
+
+        public DirectoryList ReadJpegSegments(IEnumerable<JpegSegment> segments)
+        {
+            return new List<Directory>();
```
drewnoakes: This looks strange. What's going on here?

MetadataExtractor/Formats/Jpeg/JpegMetadataReader.cs
```diff
@@ -114,7 +116,10 @@ public static DirectoryList ProcessJpegSegments([NotNull] IEnumerable<IJpegSegme
             {
                 var readerSegmentTypes = reader.SegmentTypes;
                 var readerSegments = segments.Where(s => readerSegmentTypes.Contains(s.Type));
-                directories.AddRange(reader.ReadJpegSegments(readerSegments));
+                if(reader is JpegDnlReader)
+                    ((JpegDnlReader)reader).ReadJpegSegments(readerSegments, directories);
```
drewnoakes: This can use pattern matching:

```c#
                if (reader is JpegDnlReader jpegDnlReader)
                    jpegDnlReader.ReadJpegSegments(readerSegments, directories);
```

However I'm not sure I like this pattern. I'll comment on `JpegDnlReader` explaining why.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            var reader = new SequentialByteArrayReader(segment.Bytes);
+
+            try
+            {
+                // Only set height from DNL if it's not already defined
+                if(!directory.TryGetInt32(JpegDirectory.TagImageHeight, out int i) || i == 0)
+                {
+                    directory.Set(JpegDirectory.TagImageHeight, reader.GetUInt16());
```
drewnoakes: I am not a fan of data in one 'directory' updating data in another. In this case I'd rather see a `JpegDnlDirectory`.

This approach would be more consistent with the rest of the library and not require this deviation to the common pattern, where the list of directories is passed in.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
kwhopper: This will be resolved with a new ByteArrayUtil.EqualTo method

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
kwhopper: I'm not sure what Nadahar intended for LengthBytes and ValueBytes. Given that they do a copy, maybe it was to give access to the array contents without allowing modifications.

I'll stay with the spirit of the original Java and return a copy, but will make it ToArray as you suggest.

drewnoakes: I think you're right, that it was a defensive copy to prevent external code from modifying the contents.

If/when we update the versions of .NET we support we can change these to `ReadOnlySpan<char>`, though unfortunately that's likely a way off (in a major release).

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
kwhopper: The only reason this property exists (from what I can tell) is so JpegDhtReader can add new tables and update TagNumberOfTables to the current count from outside the directory. I think I'll try making Tables private, adding a new AddTable(HuffmanTable) method, and updating the count inside of that method instead. That makes a public List property unnecessary with the only tradeoff that the value under the tag might get updated more than once.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
kwhopper: This was necessary to implement IJpegSegmentMetadataReader. But since the current implementation calls a different method the returned list should be blank.

It will become unnecessary when JpegDnlReader gets its own directory, which is the current plan.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
kwhopper: I tend to agree, and this felt a little weird to me as well.

That said - there are many cases in ExifTool where some values or procedures need to know data from other directories. I wonder if it's almost time to consider a context object like 'metadata' in Java. We will be bumping up against this as parsing complexity advances.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
drewnoakes: On the plus side the tag value should always be correct.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
drewnoakes: Agreed. Whatever solution takes place here should be consistent across all extractors though, IMO.

kwhopper:
@drewnoakes I think most of these issues have been addressed. Please review it again and let me know if there are other questions or concerns.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
```diff
@@ -42,18 +43,23 @@ public sealed class JpegDhtReader : IJpegSegmentMetadataReader
...
-
         public DirectoryList ReadJpegSegments(IEnumerable<JpegSegment> segments)
         {
-
-            var directory = new HuffmanTablesDirectory();
-            foreach (var segment in segments)
+            // This Extract structure is a little different since we only want to return one HuffmanTablesDirectory
+            // for one-to-many segments
+            var directories = new List<Directory>();
+            if (segments.Count() > 0)
```
drewnoakes: This method ends up enumerating `segments` twice. It'd be good to do it only a single time. Perhaps declare `directory` as `null` initially, then allocate lazily within the `foreach` loop and only add it after the loop if it's non-`null`.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
```diff
@@ -42,18 +43,23 @@ public sealed class JpegDhtReader : IJpegSegmentMetadataReader
...
+                foreach (var segment in segments)
+                {
+                    Extract(new SequentialByteArrayReader(segment.Bytes), directory);
+                }
+
+                directories.Add(directory);
             }
 
-            return new List<Directory> { directory };
+            return directories;
```
drewnoakes: It should be possible to avoid allocating a `List<>` in cases where no DNL segment exists.

For example:

```c#
class Directory
{
    internal static readonly DirectoryList EmptyList = new Directory[0];
}
```

Then here if `directory` is `null` just return `Directory.EmptyList`.

Note that in the general case we'd use `Array.Empty<Directory>()` but because we also target net35 that's not an option.

BTW this is a minor thing so don't worry if you don't think it's worth it.

MetadataExtractor/Formats/Jpeg/JpegDnlDirectory.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            { TagImageHeight, "Image Height" }
+        };
+
+        public JpegDnlDirectory()
+        {
+            SetDescriptor(new JpegDnlDescriptor(this));
+        }
+
+        public override string Name => "JpegDnl";
```
drewnoakes: Perhaps use name `"JPEG DNL"`. I think that's more consistent with other directory names.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
```diff
@@ -45,42 +45,28 @@ public sealed class JpegDnlReader : IJpegSegmentMetadataReader
...
+            var directory = new JpegDnlDirectory();
 
             try
             {
-                // Only set height from DNL if it's not already defined
-                if(!directory.TryGetInt32(JpegDirectory.TagImageHeight, out int i) || i == 0)
-                {
-                    directory.Set(JpegDirectory.TagImageHeight, reader.GetUInt16());
-                }
+                directory.Set(JpegDirectory.TagImageHeight, reader.GetUInt16());
```
drewnoakes: Is this setting the wrong tag?

MetadataExtractor/Formats/Jpeg/JpegMetadataReader.cs
```diff
@@ -116,8 +116,21 @@ public static DirectoryList ProcessJpegSegments([NotNull] IEnumerable<IJpegSegme
             {
                 var readerSegmentTypes = reader.SegmentTypes;
                 var readerSegments = segments.Where(s => readerSegmentTypes.Contains(s.Type));
-                if(reader is JpegDnlReader)
-                    ((JpegDnlReader)reader).ReadJpegSegments(readerSegments, directories);
+
+                // Note: JpegDnl parsing is optional. The DNL reader only defines a Height tag, 
+                //       which is redundant unless there's some problem with the regular Jpeg Height tag
```
drewnoakes: I'm interested to hear why you would like this to be optional. My feeling is that we should extract it if it's present. I don't think the DNL segment is that common in the wild.

MetadataExtractor/Util/ByteArrayUtil.cs
```diff
@@ -17,5 +17,26 @@ public static bool StartsWith([NotNull] this byte[] source, [NotNull] byte[] pat
 
             return true;
         }
+
+        public static bool EqualTo([NotNull] this byte[] source, [NotNull] byte[] compare)
```
drewnoakes: Looks good.

This is also a `[Pure]` method.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,329 @@
...
+{
+    /// <summary>Directory of tables for the DHT (Define Huffman Table(s)) segment.</summary>
+    /// <author>Nadahar</author>
+    /// <author>Kevin Mott https://github.com/kwhopper</author>
+    [SuppressMessage("ReSharper", "MemberCanBePrivate.Global")]
+    public sealed class HuffmanTablesDirectory : Directory
+    {
+        public const int TagNumberOfTables = 1;
+
+        internal static byte[] TypicalLuminanceDcLengths = {
```
drewnoakes: These field can all be `readonly`.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
```diff
@@ -0,0 +1,110 @@
+﻿#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
drewnoakes: This line does not apply to this file.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDirectory.cs
```diff
@@ -0,0 +1,329 @@
+#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
drewnoakes: This line does not apply to this file.

MetadataExtractor/Formats/Jpeg/HuffmanTablesDescriptor.cs
```diff
@@ -0,0 +1,69 @@
+#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
drewnoakes: This line does not apply to this file.

MetadataExtractor/Formats/Jpeg/JpegDnlDescriptor.cs
```diff
@@ -0,0 +1,60 @@
+#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
drewnoakes: This line does not apply to this file.

MetadataExtractor/Formats/Jpeg/JpegDnlDirectory.cs
```diff
@@ -0,0 +1,55 @@
+#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
drewnoakes: This line does not apply to this file.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
```diff
@@ -0,0 +1,72 @@
+﻿#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
drewnoakes: This line does not apply to this file.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
```diff
@@ -0,0 +1,72 @@
...
+using MetadataExtractor.IO;
+#if NET35
+using DirectoryList = System.Collections.Generic.IList<MetadataExtractor.Directory>;
+#else
+using DirectoryList = System.Collections.Generic.IReadOnlyList<MetadataExtractor.Directory>;
+#endif
+
+namespace MetadataExtractor.Formats.Jpeg
+{
+    /// <summary>Decodes JPEG DNL data, adjusting the image height with information missing from the JPEG SOFx segment.</summary>
```
drewnoakes: This comment is no longer completely true.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
kwhopper: This does "work" but not cleanly. DirectoryList needs to also be defined in the Directory class but 'using' can't be inherited so all the other DirectoryList preprocessor conditionals still must exist.

I agree this is relatively minor so I'll have it return an empty list for now. If something changes in the future to make this easier, it can be revisited.

MetadataExtractor/Formats/Jpeg/JpegMetadataReader.cs
kwhopper: Wrong sentence structure. It's optional in the sense that nothing is parsed if jpegDirectory has an image height defined already, per Nadahar's implementation. We could change that conditional if you want but there aren't any other DNL tags that get parsed currently.

MetadataExtractor/Formats/Jpeg/JpegDnlReader.cs
kwhopper: Modified the summary and added a remark.

kwhopper:
More refinements as requested. Let me know how it looks.

MetadataExtractor/Formats/Jpeg/JpegDhtReader.cs
drewnoakes: `DirectoryList` is just a local alias, so you'd only need to copy it into the `Directory` class, not all its consumers too.

MetadataExtractor/Formats/Jpeg/JpegMetadataReader.cs
drewnoakes: I would prefer if we didn't special case things here.

Logic that spans directories (such as this) belongs in the as yet undesigned feature for extracting things like dates/dimensions/etc from any of the various places they may live. Discussed further in drewnoakes/metadata-extractor#10.

drewnoakes: Remove this special case and this PR is good to merge.

I know it'll create a difference between dotnet and java versions, but we will introduce this same behaviour in the java version too.

kwhopper:
Round 4 is ready.

PR #198, opened 2019-08-24
port some GPS changes from Java to better match output and style

kwhopper:


MetadataExtractor/Formats/Exif/GpsDescriptor.cs
```diff
@@ -177,6 +242,16 @@ public string GetGpsSpeedRefDescription()
...
         }
 
+        [CanBeNull]
+        public string GetGpsSpeedDescription()
+        {
+            if (!Directory.TryGetRational(GpsDirectory.TagSpeed, out Rational value))
+                return null;
+
+            var unit = GetGpsSpeedRefDescription();
+            return string.Format("{0} {1}", value.ToDouble().ToString("0.##"), unit == null ? "unit" : unit.ToLower());
```
drewnoakes:
```suggestion
            return string.Format("{0:0.##} {1}", value.ToDouble(), unit?.ToLower() ?? "unit");
```

PR #199, opened 2019-08-25
Handle StringValue unint parsing when input is invalid (usually becau…

kwhopper:
…se of encoding)


MetadataExtractor/StringValue.cs
```diff
@@ -87,7 +87,23 @@ int IConvertible.ToInt32(IFormatProvider provider)
...
         ushort IConvertible.ToUInt16(IFormatProvider provider) => ushort.Parse(ToString());
 
-        uint IConvertible.ToUInt32(IFormatProvider provider) => uint.Parse(ToString());
+        uint IConvertible.ToUInt32(IFormatProvider provider)
+        {
+            try
+            {
+                return uint.Parse(ToString());
+            }
+            catch(Exception)
```
drewnoakes: Can we capture a more specific kind of exception here?

MetadataExtractor/StringValue.cs
```diff
@@ -87,7 +87,23 @@ int IConvertible.ToInt32(IFormatProvider provider)
...
+            {
+                ulong val = 0;
+                foreach (var b in Bytes)
+                {
+                    val = val << 8;
+                    val += b;
```
drewnoakes: I'm not this approach makes sense if there are more than four bytes (or perhaps also when fewer).

MetadataExtractor/StringValue.cs
```diff
@@ -87,7 +87,23 @@ int IConvertible.ToInt32(IFormatProvider provider)
 
         ushort IConvertible.ToUInt16(IFormatProvider provider) => ushort.Parse(ToString());
 
-        uint IConvertible.ToUInt32(IFormatProvider provider) => uint.Parse(ToString());
+        uint IConvertible.ToUInt32(IFormatProvider provider)
```
drewnoakes: Why only `uint`? Why not all the other supported integral types as well?

PR #201, opened 2019-08-25
Catch PNG chunk processing errors

kwhopper:
Good metadata is lost when PNG chunks encounter processing errors. Refactor so errors can be recorded while preserving this good metadata.

### DIFF
--- a/MetadataExtractor/Formats/Png/PngMetadataReader.cs
+++ b/MetadataExtractor/Formats/Png/PngMetadataReader.cs
@@ -84,10 +84,26 @@ namespace MetadataExtractor.Formats.Png
         [NotNull]
         public static DirectoryList ReadMetadata([NotNull] Stream stream)
         {
-            return new PngChunkReader()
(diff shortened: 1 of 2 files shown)


MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -84,10 +84,26 @@ public static DirectoryList ReadMetadata([NotNull] string filePath)
...
-                .SelectMany(ProcessChunk)
-                .ToList();
+            List<Directory> directories = null;
+
+            var chunks = new PngChunkReader().Extract(new SequentialStreamReader(stream), _desiredChunkTypes);
+
+            foreach (var chunk in chunks)
+            {
+                if(directories == null)
+                    directories = new List<Directory>();
```
drewnoakes: As an aside, I'd like to upgrade to C# 8 soon, after which we'll be able to write `directories ??= new List<Directory>()`.

PR #203, opened 2019-09-05
Use C# 8

drewnoakes:
- Adds an explicit reference to a newer compiler that supports C# 8.
- Uses various new language features:
  - Nullable reference types, replacing all `[CanBeNull]` annotations with `?` and removing all `[NotNull]` as that's now implicit (only in the main project for now)
  - Switch expressions
  - Pattern matching
  - Simple using expressions
  - Static local functions
  - Simple default expressions


kwhopper:
Just a comment - there are a lot of good changes in these, particularly with the static import.

drewnoakes:
@kwhopper thanks.

I know this forces us to deal with the new nullable reference types language feature. I've been dogfooding it for a while at work so should be able to help with any questions you or other contributors may have when using it.

PR #208, opened 2019-09-08
Add CRX record modes

dmitry-shechtman:
Based on info from https://sno.phy.queensu.ca/~phil/exiftool/TagNames/Canon.html


drewnoakes:
Thank you :)

PR #212, opened 2019-09-09
Uniform QuickTime and RIFF checkers

dmitry-shechtman:
This also restores backwards compatibility (FileType.QuickTime for unknown major brands)


drewnoakes:
LGTM

PR #215, opened 2019-09-16
Add WAV support

dmitry-shechtman:


MetadataExtractor/Formats/Wav/WavRiffHandler.cs
```diff
@@ -0,0 +1,59 @@
+#region License
+//
+// Copyright 2002-2019 Drew Noakes
+// Ported from Java to C# by Yakov Danilov for Imazen LLC in 2014
```
kwhopper: Are these comments accurate?

drewnoakes: @kwhopper thanks for pointing this out. I know I commented on these on your PRs. Rather than point it out on all PRs , I'm going to simplify the license banner at the top so it's just a single line that points to more information in the LICENSE file.

drewnoakes: @kwhopper see #216

PR #216, opened 2019-09-17
Simplify license banners

drewnoakes:
This addresses two points I disliked about the old banners:
1. They were too verbose
2. They would mention the conversion from Java for some files but not others, and this created unnecessary review friction for PRs. This information is available in `LICENSE` and in the repository's history.


kwhopper:
Much easier to deal with. Thanks!

PR #223, opened 2019-09-18
Add string timezone tags

dmitry-shechtman:


drewnoakes:
I'll run this over the image library to see how many hits we get.

drewnoakes:
They came up in images from a Sony camera.

dmitry-shechtman:
I encountered these in CR3 images, so newer Canon models seem to produce them as well.

drewnoakes:
> I encountered these in CR3 images

Can you contribute a sample image? The actual image isn't important.

dmitry-shechtman:
All the images I tested on are from https://github.com/lclevy/canon_cr3

drewnoakes:
Thanks, I've reached out to the author of those images. Hopefully they get back to me and give us permission to add them to the image library.

PR #224, opened 2019-09-18
Mimic Exif hierarchy in CRX

dmitry-shechtman:


MetadataExtractor/Formats/QuickTime/QuickTimeTiffHandler.cs
```diff
@@ -22,7 +24,11 @@ public override void SetTiffMarker(int marker)
             {
                 throw new TiffProcessingException($"Unexpected TIFF marker: 0x{marker:X}");
             }
-            PushDirectory(new T());
+            var directory = new T
+            {
+                Parent = Directories.OfType<TParent>().SingleOrDefault()
```
drewnoakes: This will throw if there's more than one instance of `TParent`, which in general happens frequently.

We need to be explicit and pass the parent to the ctor.

MetadataExtractor/Formats/QuickTime/QuickTimeTiffHandler.cs
```diff
@@ -22,7 +24,11 @@ public override void SetTiffMarker(int marker)
             {
                 throw new TiffProcessingException($"Unexpected TIFF marker: 0x{marker:X}");
             }
-            PushDirectory(new T());
+            var directory = new T
+            {
+                Parent = Directories.OfType<TParent>().FirstOrDefault()
```
drewnoakes: This is prone to selecting the wrong parent. The first one might not be the correct one. I would suggest adding the parent directory as a constructor parameter, storing it in a field, then adding it here.

You can then remove the `TParent` type parameter.

dmitry-shechtman: You mean an additional directory stack in `QucikTimeMetadataReader`? I'm not sure if adding state to the latter is better than the current solution.

drewnoakes: Isn't `SetTiffMarker` only called once? I don't think a stack is needed.

drewnoakes: I meant the state would be in this class (`QuickTimeTiffHandler`).

dmitry-shechtman: The reader encounters the atoms in order:

* CMT1
* CMT2
* CMT3
* CMT4

For each of those, a new `ITiffHandler` instance is created, meaning a directory stack created from scratch every time.

drewnoakes: Thanks for clarifying. I still feel it should be possible to pass the parent forward somehow rather than looking it up, but will have to pull down your branch to play around with it.

PR #225, opened 2019-09-18
Extract XMP from CRX

dmitry-shechtman:


drewnoakes:
(FYI I've disabled the setting that causes the "out of date with base branch" warnings.)

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -176,6 +178,22 @@ void Handler(AtomCallbackArgs a)
...
+                    {
+                        var XMP = new byte[] { 0xbe, 0x7a, 0xcf, 0xcb, 0x97, 0xa9, 0x42, 0xe8, 0x9c, 0x71, 0x99, 0x94, 0x91, 0xe3, 0xaf, 0xac };
+                        if (a.BytesLeft >= XMP.Length)
+                        {
+                            var uuid = a.Reader.GetBytes(XMP.Length);
+                            if (XMP.RegionEquals(0, XMP.Length, uuid))
+                            {
+                                var xmpBytes = a.Reader.GetNullTerminatedBytes((int)a.BytesLeft);
+                                var xmpDirectory = new XmpReader().Extract(xmpBytes);
+                                xmpDirectory.Parent = directories.OfType<ExifIfd0Directory>().SingleOrDefault();
```
drewnoakes: This will throw if there is more than one `ExifIfd0Directory`, which can happen. We should be explicit about parents, though in this case it looks like there may not be a parent directory, and that this is a top level directory.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -176,6 +178,22 @@ void Handler(AtomCallbackArgs a)
...
+                    {
+                        var XMP = new byte[] { 0xbe, 0x7a, 0xcf, 0xcb, 0x97, 0xa9, 0x42, 0xe8, 0x9c, 0x71, 0x99, 0x94, 0x91, 0xe3, 0xaf, 0xac };
+                        if (a.BytesLeft >= XMP.Length)
+                        {
+                            var uuid = a.Reader.GetBytes(XMP.Length);
+                            if (XMP.RegionEquals(0, XMP.Length, uuid))
+                            {
+                                var xmpBytes = a.Reader.GetNullTerminatedBytes((int)a.BytesLeft);
+                                var xmpDirectory = new XmpReader().Extract(xmpBytes);
+                                xmpDirectory.Parent = directories.OfType<ExifIfd0Directory>().FirstOrDefault();
```
drewnoakes: I'm still not sure this is correct. The parent here should be the metadata container that wrapped this inner collection. In this case we are in a QuickTime container which does not have a directory, so I would just leave the parent null and remove this line.

```suggestion
```

dmitry-shechtman: The idea here was to mimic the directory structure seen in CR2. I was actually considering using `Directories.Insert()`  to put `xmpDirectory` in front of its siblings.

drewnoakes: Can you help me understand the use case for needing similar structure?

dmitry-shechtman: There is not a strict *need*, just that some applications that process CR2 might find it handy.

drewnoakes:
I removed the parent directory assignment and merged. None of the other directories created in this handler have parents. That's not to say they shouldn't but I feel further thought is needed here. I wanted to get this in ahead of a release. Thanks again.

PR #227, opened 2019-10-25
QuickTime Compatible Brands value correction

KleasMcHolthuis:
The Description of the Tag QuickTime Compatible Brands is not a string, but a list. This results in  a description of "System.Collections.Generic.List`1[System.String]}"
In order to get the values of the compatible brands, you could use a string join.

[My PR did not compile, sorry about that. I should have opened an issue instead of a PR without testing]


drewnoakes:
Thanks for this. The compile error only occurred in .NET 3.5. I fixed it in 700d58e8.

PR #233, opened 2019-11-12
Add TGA support

dmitry-shechtman:


MetadataExtractor/Formats/Tga/TgaDeveloperReader.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+using MetadataExtractor.IO;
+using System.IO;
+
+namespace MetadataExtractor.Formats.Tga
+{
+    /// <summary>Reads TGA image file extension area.</summary>
+    /// <author>Dmitry Shechtman</author>
+    sealed class TgaDeveloperReader : TgaDirectoryReader<TgaDeveloperDirectory, IndexedReader>
```
drewnoakes: Please use explicit accessibilities. In general these inner-working types are made public in this library.

MetadataExtractor.sln
```diff
@@ -76,6 +76,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Specs", "Specs", "{1806EC0F
 		wiki\docs\PCX File Format.html = wiki\docs\PCX File Format.html
 		wiki\docs\PCX File Format.txt = wiki\docs\PCX File Format.txt
 		wiki\docs\QTFF.pdf = wiki\docs\QTFF.pdf
+		wiki\docs\tga_specs.pdf = wiki\docs\tga_specs.pdf
```
drewnoakes: Were you able to push to the wiki repo and update the submodule in this PR?

MetadataExtractor/Formats/Tga/TgaDeveloperReader.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+using MetadataExtractor.IO;
+using System.IO;
+
+namespace MetadataExtractor.Formats.Tga
+{
+    /// <summary>Reads TGA image file extension area.</summary>
```
drewnoakes: Copy/paste error?

MetadataExtractor/Formats/Tga/TgaFooterReader.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// <summary>Reads TGA image file footer.</summary>
+    /// <author>Dmitry Shechtman</author>
+    sealed class TgaFooterReader : TgaReader<TgaFooter, SequentialReader>
+    {
+        private const int FooterSize = 26;
+
+        private static readonly byte[] FooterMagic = Encoding.ASCII.GetBytes("TRUEVISION-XFILE.\0");
+
+        public static readonly TgaFooterReader Instance = new TgaFooterReader();
```
drewnoakes: Why did you make these singletons?

MetadataExtractor/Formats/Tga/TgaFooterReader.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.IO;
+using System.Text;
+
+namespace MetadataExtractor.Formats.Tga
+{
+    struct TgaFooter
+    {
+        public int extOffset;
+        public int devOffset;
+        public byte[] signature;
```
drewnoakes: Please capitalise member names.

MetadataExtractor/Formats/Tga/TgaHeaderReader.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        public const int HeaderSize = 18;
+
+        public static readonly TgaHeaderReader Instance = new TgaHeaderReader();
+
+        private TgaHeaderReader()
+        {
+        }
+
+        public bool TryExtract(byte[] buffer, out TgaHeaderDirectory directory)
```
drewnoakes: Please set `directory` to `null` when returning `false`. Consuming code might one day reasonably expect or depend upon that.

MetadataExtractor/Util/FileType.cs
```diff
@@ -160,6 +165,7 @@ public static class FileTypeExtensions
             null,
             "video/quicktime",
             "image/x-portable-graymap",
+            null,
             null
```
drewnoakes:
```suggestion
            "image/x-targa"
```

MetadataExtractor/Util/FileTypeDetector.cs
```diff
@@ -94,7 +103,7 @@ public static FileType DetectFileType(Stream stream)
             if (!stream.CanSeek)
                 throw new ArgumentException("Must support seek", nameof(stream));
 
-            var maxByteCount = _root.MaxDepth;
+            var maxByteCount = Math.Max(_root.MaxDepth, TgaHeaderReader.HeaderSize);
```
drewnoakes: We should change `_fixedCheckers` to be an enumerable of value tuples with an additional `int? HeaderSize`, then scan all checkers for their max non-null size for use here. It'll force us to think about this requirement for the other existing checkers and future ones too.

dmitry-shechtman:
> I always thought of these files as "Targa" files. Is "TGA" a more common name, or just the file extension? In general I've tried to use the friendly name in type names, directory `Name` values, etc.

It is commonly known as TARGA, but that's the name of the video card it was used for. I stuck to the official name.

MetadataExtractor/Formats/Tga/TgaDeveloperReader.cs
dmitry-shechtman: Should they really be public?

MetadataExtractor.sln
dmitry-shechtman: Sorry about that. I used VS to add the solution item, but I didn't find a way to push it.

MetadataExtractor/Formats/Tga/TgaFooterReader.cs
dmitry-shechtman: These classes don't have states, but have virtual members, so they can't be made static.

MetadataExtractor/Util/FileTypeDetector.cs
dmitry-shechtman: Maybe extract an interface so the check method and the size value are closer together?

MetadataExtractor/Formats/Tga/TgaFooterReader.cs
drewnoakes: Why not new them up each time? There doesn't seem much benefit to keeping long-lived instances as far as I can tell.

MetadataExtractor.sln
drewnoakes: The wiki repo is added as a git submodule to this repo, so that it can appear in VS. If you add or change files in it, they must first be pushed to the submodule repo, then this PR can update its submodule pointer (commit) to the new commit there. I think you should have push access to the wiki repo. If not, let me know where I can download this spec and I'll be sure to add it.

MetadataExtractor/Formats/Tga/TgaDeveloperReader.cs
drewnoakes: Maybe. Pros and cons either way. I don't feel strongly about them being public in this instance.

For reference, often people ask how to do specific things with a format and the answer is usually to use the classes like this for their purposes. In this case they appear to be a fixed set of internal processing types, so they probably don't need to be public.

MetadataExtractor/Util/FileTypeDetector.cs
drewnoakes: Interface works.

drewnoakes:
Do you have some sample `.tga` files you can provide for the sample library to help prevent regressions here?

MetadataExtractor.sln
dmitry-shechtman: I tried `git push origin HEAD:master`, but got Permission denied. I got the spec from http://tfc.duke.free.fr/coding/tga_specs.pdf

MetadataExtractor/Formats/Tga/TgaFooterReader.cs
dmitry-shechtman: Done.

MetadataExtractor/Util/FileTypeDetector.cs
dmitry-shechtman: Done.

dmitry-shechtman:
> Do you have some sample `.tga` files you can provide for the sample library to help prevent regressions here?

Some of the TGAs I tested on are from ezQuake:

https://github.com/ezQuake/ezquake-media/tree/master/game/textures

Specific files that might be useful:

models/simple_backpack_0.tga - top to bottom, 0 attribute bits per pixel
shellmap.tga - uncompressed true color

Both have footers, but no extension/developer areas.

MetadataExtractor.sln
drewnoakes: No worries. I pushed the file to the wiki. We can update the submodule later, or feel free to do so.

drewnoakes:
Thanks! I fixed the merge conflicts and made a few other tweaks, verifying the output matches for the ezQuake sample images.

PR #235, opened 2019-11-17
Port of Java PR #267 - Clipping Path Name / Path Information Extraction

kwhopper:
port of Java PR drewnoakes/metadata-extractor#267

### DIFF
--- a/MetadataExtractor/Formats/Photoshop/Knot.cs
+++ b/MetadataExtractor/Formats/Photoshop/Knot.cs
@@ -13,7 +13,7 @@ namespace MetadataExtractor.Formats.Photoshop
     /// <author>Kevin Mott https://github.com/kwhopper</author>
     public class Knot
     {
-        private double[] p_points = new double[6];
+        private readonly double[] _points = new double[6];
 
         public Knot(string type)
         {
@@ -21,29 +21,21 @@ namespace MetadataExtractor.Formats.Photoshop
         }
 
         /// <summary>
-        /// Add an individual coordinate value (x or y) to
+        /// Add/Get an individual coordinate value (x or y) to/from
         /// points array (6 points per knot)
-        /// </summary>
-        /// <param name="index">location of point to be added in points</param>
-        /// <param name="point">coordinate value to be added to points</param>
-        public void SetPoint(int index, double point)
-        {
-            p_points[index] = point;
-        }
-
-        /// <summary>
-        /// Get an individual coordinate value (x or y)
         /// </summary>
         /// <param name="index"></param>
         /// <returns>an individual coordinate value</returns>
-        public double GetPoint(int index)
+        /// <remarks>Define the indexer to allow client code to use [] notation</remarks>
+        public double this[int index]
         {
-            return p_points[index];
+            get => _points[index];
+            set => _points[index] = value;
         }
 
         /// <summary>
         /// Get the type of knot (linked or unlinked)
         /// </summary>
-        public string Type { get; private set; }
+        public string Type { get; }
     }
 }
--- a/MetadataExtractor/Formats/Photoshop/Knot.cs
+++ b/MetadataExtractor/Formats/Photoshop/Knot.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace MetadataExtractor.Formats.Photoshop
+{
+    /// <summary>
+    /// Represents a knot created by Photoshop:
+    /// <list type="bullet">
+    /// <item>Linked knot</item>
+    /// <item>Unlinked knot</item>
+    /// </list>
+    /// </summary>
+    /// <author>Payton Garland</author>
+    /// <author>Kevin Mott https://github.com/kwhopper</author>
+    public class Knot
+    {
+        private double[] p_points = new double[6];
+
+        public Knot(string type)
+        {
+            Type = type;
+        }
+
+        /// <summary>
+        /// Add an individual coordinate value (x or y) to
+        /// points array (6 points per knot)
+        /// </summary>
+        /// <param name="index">location of point to be added in points</param>
+        /// <param name="point">coordinate value to be added to points</param>
+        public void SetPoint(int index, double point)
+        {
+            p_points[index] = point;
+        }
+
+        /// <summary>
+        /// Get an individual coordinate value (x or y)
+        /// </summary>
+        /// <param name="index"></param>
+        /// <returns>an individual coordinate value</returns>
+        public double GetPoint(int index)
+        {
+            return p_points[index];
+        }
+
+        /// <summary>
+        /// Get the type of knot (linked or unlinked)
+        /// </summary>
+        public string Type { get; private set; }
+    }
+}
--- a/MetadataExtractor/Formats/Photoshop/PhotoshopDescriptor.cs
+++ b/MetadataExtractor/Formats/Photoshop/PhotoshopDescriptor.cs
@@ -1,5 +1,7 @@
 // Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
+using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Text;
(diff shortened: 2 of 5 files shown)


MetadataExtractor/Formats/Photoshop/Knot.cs
```diff
@@ -0,0 +1,49 @@
...
+            Type = type;
+        }
+
+        /// <summary>
+        /// Add an individual coordinate value (x or y) to
+        /// points array (6 points per knot)
+        /// </summary>
+        /// <param name="index">location of point to be added in points</param>
+        /// <param name="point">coordinate value to be added to points</param>
+        public void SetPoint(int index, double point)
```
drewnoakes: These get and set methods could be replaces with an indexer property if you liked (not available in Java). Up to you.

MetadataExtractor/Formats/Photoshop/Knot.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <returns>an individual coordinate value</returns>
+        public double GetPoint(int index)
+        {
+            return p_points[index];
+        }
+
+        /// <summary>
+        /// Get the type of knot (linked or unlinked)
+        /// </summary>
+        public string Type { get; private set; }
```
drewnoakes:
```suggestion
        public string Type { get; }
```

MetadataExtractor/Formats/Photoshop/PhotoshopDescriptor.cs
```diff
@@ -47,7 +49,11 @@ public PhotoshopDescriptor(PhotoshopDirectory directory)
                     return GetPrintScaleDescription();
                 case PhotoshopDirectory.TagPixelAspectRatio:
                     return GetPixelAspectRatioString();
+                case PhotoshopDirectory.TagClippingPathName:
+                    return GetClippingPathNameString(tagType);
                 default:
+                    if (tagType >= 0x07D0 && tagType <= 0x0BB6)
```
drewnoakes: What are these two numbers? Can they be defined as constants somewhere?

MetadataExtractor/Formats/Photoshop/PhotoshopReader.cs
```diff
@@ -133,7 +140,23 @@ public DirectoryList Extract(SequentialReader reader, int length)
                             directories.Add(xmpDirectory);
                             break;
                         default:
-                            directory.Set(tagType, tagBytes);
+                            if (tagType >= 0x07D0 && tagType <= 0x0BB6)
```
drewnoakes: Ditto.

MetadataExtractor/Formats/Photoshop/Subpath.cs
```diff
@@ -0,0 +1,56 @@
...
+        public void Add(Knot knot)
+        {
+            p_knots.Add(knot);
+        }
+
+        /// <summary>
+        /// Gets size of knots list
+        /// </summary>
+        /// <returns>size of knots List</returns>
+        public int Size()
```
drewnoakes: This could be a property. Alternatively, remove this `Size` member, make `GetKnots` an `IReadOnlyList<Knot>` and allow callers to use `subPath.Knots.Count`, which would feel more idiomatic in .NET. The name "size" is more Java-esque.

MetadataExtractor/Formats/Photoshop/Subpath.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <author>Kevin Mott https://github.com/kwhopper</author>
+    public class Subpath
+    {
+        private List<Knot> p_knots = new List<Knot>();
+
+        public Subpath() : this("")
+        {
+        }
+
+        public Subpath(string type)
```
drewnoakes: You can merge these two constructors and make the parameter optional, eg:

```suggestion
        public Subpath(string type = "")
```

MetadataExtractor/Formats/Photoshop/Subpath.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            return p_knots.Count;
+        }
+
+        public IEnumerable<Knot> GetKnots()
+        {
+            return p_knots;
+        }
+
+        public string Type { get; private set; }
```
drewnoakes:
```suggestion
        public string Type { get; }
```

MetadataExtractor/Formats/Photoshop/Subpath.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <item>Closed Bezier knot, unlinked</item>
+    /// <item>Open Bezier knot, linked</item>
+    /// <item>Open Bezier knot, unlinked</item>
+    /// </list>
+    /// </summary>
+    /// <author>Payton Garland</author>
+    /// <author>Kevin Mott https://github.com/kwhopper</author>
+    public class Subpath
+    {
+        private List<Knot> p_knots = new List<Knot>();
```
drewnoakes:
```suggestion
        private readonly List<Knot> p_knots = new List<Knot>();
```

MetadataExtractor/Formats/Photoshop/PhotoshopReader.cs
```diff
@@ -133,7 +140,23 @@ public DirectoryList Extract(SequentialReader reader, int length)
...
+                                // Append description(name) to end of byte array with 1 byte before the description representing the length
+                                for (int i = tagBytes.Length - description.Length - 1; i < tagBytes.Length; i++)
+                                {
+                                    if (i % (tagBytes.Length - description.Length - 1 + description.Length) == 0)
+                                        tagBytes[i] = (byte)description.Length;
+                                    else
+                                        tagBytes[i] = (byte)description[i - (tagBytes.Length - description.Length - 1)];
+                                }
+                                PhotoshopDirectory.TagNameMap[0x07CF + clippingPathCount] = "Path Info " + clippingPathCount;
+                                directory.Set(0x07CF + clippingPathCount, tagBytes);
```
drewnoakes: Ditto, constants.

MetadataExtractor/Formats/Photoshop/PhotoshopReader.cs
```diff
@@ -71,16 +72,22 @@ public DirectoryList Extract(SequentialReader reader, int length)
                     pos += 2;
 
                     // A variable number of bytes holding a pascal string (two leading bytes for length).
-                    var descriptionLength = reader.GetByte();
+                    int descriptionLength = reader.GetByte();
                     pos += 1;
 
                     // Some basic bounds checking
-                    if (descriptionLength + pos > length)
+                    if (descriptionLength < 0 || descriptionLength + pos > length)
```
drewnoakes: Doesn't `GetByte` return a `byte`, which cannot be negative?

MetadataExtractor/Formats/Photoshop/Subpath.cs
kwhopper: That works, until NET3.5 compat kicks in - then it gets a little weird. I might change the name from GetKnots to Knots and "Size" to "KnotsCount". Otherwise, it has to be the Count() Linq method, which will cause an enumeration. I'll post it and see what you think.

kwhopper:
Tried to accommodate all your suggestions. Thanks

MetadataExtractor/Formats/Photoshop/PhotoshopReader.cs
```diff
@@ -71,16 +72,22 @@ public DirectoryList Extract(SequentialReader reader, int length)
                     pos += 2;
 
                     // A variable number of bytes holding a pascal string (two leading bytes for length).
-                    var descriptionLength = reader.GetByte();
+                    int descriptionLength = reader.GetByte();
```
drewnoakes:
```suggestion
                    byte descriptionLength = reader.GetByte();
```

drewnoakes: I'd use the actual type here rather that widening to a signed type. It'll prevent wondering whether it may be less than zero below.

kwhopper:
Removed the int cast of descriptionLength. It initially had to be int because the Java code adds pos to it before entering the loop so it was either start with an int or cast pos. I removed that logic and used a decrement of descriptionLength in the loop instead. Also fixed a regression when using the new clippingpath constants.

If this looks ok to you, I'll go ahead and merge. Thanks!

drewnoakes:
Thanks again.

PR #236, opened 2019-11-28
port EPS support from Java

kwhopper:
Self-explanatory.

One debatable addition is the StringBuilderExtensions class. I tried to keep the IndexOf portion since it saves some Strings, and adapted a simple set of methods from the internet. The URL is in the comments. If you'd rather not use this, that's fine but it will need to ToString().IndexOf instead.


MetadataExtractor/Formats/Eps/EpsReader.cs
```diff
@@ -0,0 +1,420 @@
...
+
+                // Stop when we hit a line that is not a comment
+                if (line.Length != 0 && line[0] != '%')
+                    break;
+
+                string name;
+
+                // ':' signifies there is an associated keyword (should be put in directory)
+                // otherwise, the name could be a marker
+                int colonIndex = line.IndexOf(":");
```
drewnoakes: Is this the only use of `IndexOf`? If so, you can make the implementation simpler by searching for a char rather than a string.

kwhopper: Most of these are addressed. Let me know if there are any other issues. Thanks

kwhopper:
Most of these are addressed. Let me know if there are any other issues. Thanks

PR #238, opened 2019-11-30
make readonly subdir string instead of explicit path strings

kwhopper:


drewnoakes:
I've started moving this code into the images repo BTW.

https://github.com/drewnoakes/metadata-extractor-images/tree/source/src

PR #240, opened 2019-12-11
192, 234 - Add Support for Quicktime metadata incl GPS and creationDate

d5smith:
Does not currently support Optional  Item_info atom or Optional name atom inside the the Value Atom as described in the spec as none of the sample data I had access to had these attributes.


MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -22,6 +23,9 @@ public static class QuickTimeMetadataReader
     {
         private static readonly DateTime _epoch = new DateTime(1904, 1, 1);
 
+        private static uint entryCount = 0;
+        private static readonly List<string> metaDataKeys = new List<string>();
```
drewnoakes: These cannot be static as it will not be safe for concurrent access.

Fixing this might require adding a data object to the `QuickTimeReader.ProcessAtoms` API so we can flow such data around. There may be an easier way of fixing this however.

MetadataExtractor.Tests/Formats/Mov/MovMetadataReaderTest.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace MetadataExtractor.Tests.Formats.Mov
+{
+    /// <summary>Unit tests for <see cref="QuickTimeMetadataReader"/>.</summary>
+    public sealed class MovMetadataReaderTest
+    {
+
+        [Fact]
+        public void ExtractMetadataUsingStream()
+        {
+            using var stream = TestDataUtil.OpenRead("Data/with-gps.mov");
```
drewnoakes: Thank you for writing tests. On other repos I find not enough PRs have tests! However in this project, because of the sheer number of data formats required, instead of adding media to the source repository we use a separate media library which tracks all output for all files within, across both the .NET and Java implementations.

Could you please remove the `.mov` file and unit tests from this PR? You'll need to edit your PR branch's history. Let me know if I can help with that.

Instead, let's add the file to this repo:

https://github.com/drewnoakes/metadata-extractor-images

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataHeaderDirectory.cs
```diff
@@ -0,0 +1,77 @@
...
+            { TagGenre,             "com.apple.quicktime.genre"},
+            { TagInformation,       "com.apple.quicktime.information"},
+            { TagKeywords,          "com.apple.quicktime.keywords"},
+            { TagLocationIso6709,   "com.apple.quicktime.location.ISO6709"},
+            { TagProducer,          "com.apple.quicktime.producer"},
+            { TagPublisher,         "com.apple.quicktime.publisher"},
+            { TagSoftware,          "com.apple.quicktime.software"},
+            { TagYear,              "com.apple.quicktime.year"},
+            { TagCollectionUser,    "com.apple.quicktime.collection.user"},
+            { TagRatingUser,        "com.apple.quicktime.rating.user"}
```
drewnoakes: Can we have a different map for user friendly names for these tags? Other directories' tags have names like "User Rating" rather than "com.apple.quicktime.rating.user".

MetadataExtractor.Tests/Formats/Mov/MovMetadataReaderTest.cs
drewnoakes: Looks like that file is already there: https://github.com/drewnoakes/metadata-extractor-images/commit/a5895dc0b5b3b47df740bffea5451b8e951d26ad

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
d5smith: Yup, I called that out in the issue discussion thread before creating the PR.
Adding a data object `QuickTimeReader.ProcessAtoms` may be the easiest solution. 
Initially I did not want to mess with / clutter that, but I think it will work. I will update the PR. Thanks.

MetadataExtractor.Tests/Formats/Mov/MovMetadataReaderTest.cs
d5smith: Yeah, that's where I got it from. it was referenced in the issue thread.
Yup, I'll remove it from my branch's history.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataHeaderDirectory.cs
d5smith: I will put friendly names in the values of the `_tagNameMap `.
I will statically create the `_nameTagMap` map then to explicitly map the string keys to the tag keys.
This is required because unlike other Atoms, not all values are present, values are not orders and they are identified by their string keys.

d5smith: I will put friendly names in the values of the _tagNameMap .
I will statically create the _nameTagMap map then to explicitly map the string keys to the tag keys.
This is required because unlike other Atoms, not all values are present, values are not orders and they are identified by their string keys.

d5smith:
I just created a separate pull request to replace this since I was working from a different env so I am closing this one. thanks.

PR #241, opened 2019-12-11
Add Support for Quicktime metadata incl GPS and CreationDate

d5smith:
Does not support Optional Item_info atom or Optional name atom inside the the Value Atom as described in the spec as none of the sample data I had access to had these attributes.

### DIFF
--- a/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
+++ b/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using MetadataExtractor.Formats.Exif;
 using MetadataExtractor.Formats.Exif.Makernotes;
 using MetadataExtractor.Formats.Tiff;
@@ -25,6 +26,7 @@ namespace MetadataExtractor.Formats.QuickTime
         public static DirectoryList ReadMetadata(Stream stream)
         {
             var directories = new List<Directory>();
+            var metaDataKeys = new List<string>();
 
             QuickTimeReader.ProcessAtoms(stream, Handler);
 
@@ -113,6 +115,59 @@ namespace MetadataExtractor.Formats.QuickTime
                 }
             }
 
+            void MetaDataHandler(AtomCallbackArgs a)
+            {
+                // see https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/Metadata/Metadata.html
+                switch (a.TypeString)
+                {
+                    case "keys":
+                    {
+                        var version = a.Reader.GetByte();
+                        var flags = a.Reader.GetBytes(3);
+                        var entryCount = a.Reader.GetUInt32();
+                        for (int i = 1; i <= entryCount; i++)
+                        {
+                            var keySize = a.Reader.GetUInt32();
+                            var keyValueSize = (int)keySize - 8;
+                            var keyNamespace = a.Reader.GetUInt32();
+                            var keyValue = a.Reader.GetBytes(keyValueSize);
+                            metaDataKeys.Add(Encoding.UTF8.GetString(keyValue));
+                        }
+                        break;
+                    }
+                    case "ilst":
+                    {
+                        var directory = new QuickTimeMetadataHeaderDirectory();
+                        // Iterate over the list of Metadata Item Atoms.
+                        for (int i = 0; i < metaDataKeys.Count; i++)
+                        {
+                            long atomSize = a.Reader.GetUInt32();
+                            var atomType = a.Reader.GetUInt32();
+
+                            // Indexes into the metadata item keys atom are 1-based (1…entry_count).
+                            // atom type for each metadata item atom should be set equal to the index of the key
+                            var key = metaDataKeys[(int)atomType - 1];
+
+                            // Value Atom
+                            var typeIndicator = a.Reader.GetBytes(4);
+                            var localeIndicator = a.Reader.GetBytes(4);
+
+                            // Data Atom
+                            var dataTypeIndicator = a.Reader.GetBytes(4);
+                            var dataLocaleIndicator = a.Reader.GetBytes(4);
+                            var data = a.Reader.GetBytes((int)atomSize - 24);
+                            var dataStr = Encoding.UTF8.GetString(data);
+                            if (directory.TryGetTag(key, out int tag))
+                            {
+                                directory.Set(tag, dataStr);
+                            }
+                        }
+                        directories.Add(directory);
+                        break;
+                }
+            }
+            }
+
             void MoovHandler(AtomCallbackArgs a)
             {
                 switch (a.TypeString)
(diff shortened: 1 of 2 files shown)


MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -113,6 +115,59 @@ void UuidHandler(AtomCallbackArgs a)
...
+                        var directory = new QuickTimeMetadataHeaderDirectory();
+                        // Iterate over the list of Metadata Item Atoms.
+                        for (int i = 0; i < metaDataKeys.Count; i++)
+                        {
+                            long atomSize = a.Reader.GetUInt32();
+                            var atomType = a.Reader.GetUInt32();
+
+                            // Indexes into the metadata item keys atom are 1-based (1�entry_count).
+                            // atom type for each metadata item atom should be set equal to the index of the key
+                            var key = metaDataKeys[(int)atomType - 1];
```
drewnoakes: Please do a bounds check here and, if out of bounds, add an error to the directory rather than throwing an exception.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -113,6 +115,59 @@ void UuidHandler(AtomCallbackArgs a)
...
+
+                            // Value Atom
+                            var typeIndicator = a.Reader.GetBytes(4);
+                            var localeIndicator = a.Reader.GetBytes(4);
+
+                            // Data Atom
+                            var dataTypeIndicator = a.Reader.GetBytes(4);
+                            var dataLocaleIndicator = a.Reader.GetBytes(4);
+                            var data = a.Reader.GetBytes((int)atomSize - 24);
```
drewnoakes: Again, please validate that `atomSize - 24` is not negative to prevent an exception, adding an error to the dictionary if needed.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -142,10 +142,22 @@ void MetaDataHandler(AtomCallbackArgs a)
...
+                            }
                             var atomType = a.Reader.GetUInt32();
 
                             // Indexes into the metadata item keys atom are 1-based (1�entry_count).
-                            // atom type for each metadata item atom should be set equal to the index of the key
+                            // atom type for each metadata item atom is the index of the key
+                            if (atomType < 1 || atomType > metaDataKeys.Count)
+                            {
+                                directory.AddError("Invalid ilist AtomType");
+                                directories.Add(directory);
```
drewnoakes: Won't this result in `directory` being added to `directories` twice?

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -142,10 +142,22 @@ void MetaDataHandler(AtomCallbackArgs a)
                         for (int i = 0; i < metaDataKeys.Count; i++)
                         {
                             long atomSize = a.Reader.GetUInt32();
+                            if (atomSize < 24)
+                            {
+                                directory.AddError("Invalid ilist AtomSize");
+                                directories.Add(directory);
```
drewnoakes: Ditto.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -113,6 +115,69 @@ void UuidHandler(AtomCallbackArgs a)
...
+                            }
+                            var key = metaDataKeys[(int)atomType - 1];
+
+                            // Value Atom
+                            var typeIndicator = a.Reader.GetBytes(4);
+                            var localeIndicator = a.Reader.GetBytes(4);
+
+                            // Data Atom
+                            var dataTypeIndicator = a.Reader.GetBytes(4);
+                            var dataLocaleIndicator = a.Reader.GetBytes(4);
```
drewnoakes: Looking at the spec, it seems that these values cannot be ignored. The value is not guaranteed to be a UTF-8 string. The type indicator is packed to contain a [well-known type](https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/Metadata/Metadata.html#//apple_ref/doc/uid/TP40000939-CH1-SW28) field which instructs the encoding to use for the data value (e.g. UTF-8, UTF-16, etc...).

Consider reading `int32` values rather than `byte[]` as the former will not cause heap allocations.

We don't have any built-in support for per-value locale support. It may be interesting to add, though this would be the only usage I know of currently.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -113,6 +115,69 @@ void UuidHandler(AtomCallbackArgs a)
...
+                            var data = a.Reader.GetBytes((int)atomSize - 24);
+                            var dataStr = Encoding.UTF8.GetString(data);
```
drewnoakes: This should use [`StringValue`](https://github.com/drewnoakes/metadata-extractor-dotnet/blob/master/MetadataExtractor/StringValue.cs) with the encoding defined by the type indicator. This allows a consumer to choose a different encoding at some later point. We regularly see data where the advertised encoding is incorrect, and `StringValue` gives users a way forward.

d5smith: I will do so

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
d5smith: I took some of my assumptions from [Table 3-6](https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/Metadata/Metadata.html#//apple_ref/doc/uid/TP40000939-CH1-SW43) (though this would leave `Artwork `and `UserRating` improperly handled).
Are you looking for a full implementation of `Well-Known Types` or would you be okay with providing support for types 1, 13, 14, 23 and 27 and treating other values as errors?

As for locale, I was hoping to provide an incremental improvement by adding support for GPS location retrieval, All of the samples that I had access to had this value set to `\0\0` so I did not have a need for it at the time.

drewnoakes: > would you be okay with providing support for types 1, 13, 14, 23 and 27 and treating other values as errors?

Yes that's a long table and I don't think we need to handle everything. Adding the framework to support different encodings and covering the ones we think would be most common and easiest is great for now. Having even basic support now lowers the bar for others who stumble upon data that's not covered.

I would avoid treating unsupported encodings as errors. Maybe just store the raw bytes instead. That's more in line with what's done elsewhere.

drewnoakes:
Thanks for this and apologies it took a while to get to it. I pulled the code down and played with it a bit. Very nice!

Would love to have more QuickTime files for testing against. If you have any you could share, please do so.

d5smith:
Glad it worked out. I don't have any more examples that I can share atm, but if I find good candidates i will send them along. Thanks. Cheers.

Issue #196, opened 2019-07-25
Test against images generated by fuzzing


drewnoakes:
In https://github.com/drewnoakes/metadata-extractor/issues/422 @alpire generously provided a set of images that were generated by fuzzing which trigger various exceptions that indicate potential bugs in the implementation when dealing with malformed inputs (e.g. divide by zero, bounds exceptions, negative array lengths and so forth).

We should run those images against the .NET implementation too and fix any resulting issues.

(When fixing these, let's use one pull request per problem.)

patricksadowski:
There is one remaining problem. The new files in the _mp3_ directory throw an `ImageProcessingException` with the message _"File format could not be determined"_. But I think that is fine.

Here is the output of the new files generated by fuzzing:
```
C:\projects\drewnoakes\metadata-extractor-images\heic/IllegalArgumentException.HeifReader.processBoxes.heif
C:\projects\drewnoakes\metadata-extractor-images\heic/NegativeArraySizeException.HeifReader.processBoxes.heif
C:\projects\drewnoakes\metadata-extractor-images\heic/NegativeArraySizeException.ItemInfoBox.init.heif
C:\projects\drewnoakes\metadata-extractor-images\heic/NullPointerException.HeifPictureHandler.processBox.heif
C:\projects\drewnoakes\metadata-extractor-images\jpg/NegativeArraySizeException.DuckyReader.extract.mp3
C:\projects\drewnoakes\metadata-extractor-images\jpg/NegativeArraySizeException.DuckyReader.extract.mp3
        [Ducky] End of data reached.
        [Ducky] End of data reached.
        [Ducky] End of data reached.
        [Ducky] End of data reached.
        [Ducky] End of data reached.
        [Ducky] End of data reached.
        [Ducky] End of data reached.
C:\projects\drewnoakes\metadata-extractor-images\mp3/ArithmeticException.Mp3Reader.extract.mp3
        [ImageProcessingException] C:\projects\drewnoakes\metadata-extractor-images\mp3/ArithmeticException.Mp3Reader.extract.mp3
C:\projects\drewnoakes\metadata-extractor-images\mp3/ArrayIndexOutOfBoundsException.Mp3Reader.extract.mp3
        [ImageProcessingException] C:\projects\drewnoakes\metadata-extractor-images\mp3/ArrayIndexOutOfBoundsException.Mp3Reader.extract.mp3
[...]

PR #243, opened 2019-12-28
Prevent negative length in "Ducky" data

patricksadowski:
Fixes the problem with [`jpg/NegativeArraySizeException.DuckyReader.extract.mp3`](https://github.com/drewnoakes/metadata-extractor-images/commit/4244161122711b3cfdc0a317b2c699cfd94ba867#diff-ed6cbf568225fbeb5fb842e85094e967) from #196

### DIFF
--- a/MetadataExtractor/Formats/Photoshop/DuckyReader.cs
+++ b/MetadataExtractor/Formats/Photoshop/DuckyReader.cs
@@ -67,7 +67,13 @@ namespace MetadataExtractor.Formats.Photoshop
                         case DuckyDirectory.TagCopyright:
                         {
                             reader.Skip(4);
-                            directory.Set(tag, reader.GetString(length - 4, Encoding.BigEndianUnicode));
+                            length -= 4;
+                            if (length < 0)
+                            {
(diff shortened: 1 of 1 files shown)


MetadataExtractor/Formats/Photoshop/DuckyReader.cs
```diff
@@ -67,7 +67,10 @@ public DuckyDirectory Extract(SequentialReader reader)
                         case DuckyDirectory.TagCopyright:
                         {
                             reader.Skip(4);
-                            directory.Set(tag, reader.GetString(length - 4, Encoding.BigEndianUnicode));
+                            length -= 4;
+                            if (length < 0)
+                                length = 0;
+                            directory.Set(tag, reader.GetString(length, Encoding.BigEndianUnicode));
```
drewnoakes: This change definitely fixes the bug. I would think that if we see an invalid length here we should add an error and stop processing rather than continuing (similar to lines 60-61).

patricksadowski: I changed the code according to your suggestion.

PR #262, opened 2020-07-15
Update QuickTimeMetadataReader.cs

kirk-marple:
We've found that MetadataExtractor isn't seeing GPS location in Android recorded video files.

I found that GPS is stored in 'udta' atom in these files, and parsing it out into metadata solved our issue.

### DIFF
--- a/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
+++ b/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
@@ -120,6 +120,23 @@ namespace MetadataExtractor.Formats.QuickTime
                 }
             }
 
+            void UserDataHandler(AtomCallbackArgs a)
+            {
+                switch (a.TypeString)
+                {
+                    case "?xyz":
+                        var directory = new QuickTimeMetadataHeaderDirectory();
+                        var stringSize = a.Reader.GetUInt16();
+                        var languageCode = a.Reader.GetUInt16();
+                        var stringValue = a.Reader.GetBytes(stringSize);
+                        var s = Encoding.UTF8.GetString(stringValue);
+
+                        directory.Set(QuickTimeMetadataHeaderDirectory.TagGpsLocation, s);
+                        directories.Add(directory);
+                        break;
+                }
+            }
+
             void MetaDataHandler(AtomCallbackArgs a)
             {
                 // see https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/Metadata/Metadata.html
@@ -247,6 +264,11 @@ namespace MetadataExtractor.Formats.QuickTime
                     case "meta":
                     {
                         QuickTimeReader.ProcessAtoms(stream, MetaDataHandler, a.BytesLeft);
+                        break;
+                    }
+                    case "udta":
+                    {
+                        QuickTimeReader.ProcessAtoms(stream, UserDataHandler, a.BytesLeft);
                         break;
                     }
                     /*


MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -120,6 +120,23 @@ void UuidHandler(AtomCallbackArgs a)
...
+                switch (a.TypeString)
+                {
+                    case "?xyz":
+                        var directory = new QuickTimeMetadataHeaderDirectory();
+                        var stringSize = a.Reader.GetUInt16();
+                        var languageCode = a.Reader.GetUInt16();
+                        var stringValue = a.Reader.GetBytes(stringSize);
+                        var s = Encoding.UTF8.GetString(stringValue);
+
+                        directory.Set(QuickTimeMetadataHeaderDirectory.TagGpsLocation, s);
```
drewnoakes, line 134: Can this string be parsed into a `GeoLocation`? It seems the value can have three components. Is the third altitude?

Even if not prased, the string should be stored as a `StringValue` so that consumers may reinterpret the raw bytes using a different encoding. I know that seems unlikely here, but it's our standard approach for storing parsed string values and gives people an escape hatch when they have non-standards-compliant data.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -120,6 +120,23 @@ void UuidHandler(AtomCallbackArgs a)
...
             }
 
+            void UserDataHandler(AtomCallbackArgs a)
+            {
+                switch (a.TypeString)
+                {
+                    case "?xyz":
+                        var directory = new QuickTimeMetadataHeaderDirectory();
+                        var stringSize = a.Reader.GetUInt16();
+                        var languageCode = a.Reader.GetUInt16();
```
drewnoakes, line 130: Is this code interesting? If not, maybe `Skip(2)` instead (with a comment "skip language code" or similar).

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
```diff
@@ -120,6 +120,23 @@ void UuidHandler(AtomCallbackArgs a)
                 }
             }
 
+            void UserDataHandler(AtomCallbackArgs a)
+            {
+                switch (a.TypeString)
+                {
+                    case "?xyz":
+                        var directory = new QuickTimeMetadataHeaderDirectory();
```
drewnoakes, line 128: I think this should be a new directory type, eg. `QuickTimeGeoLocationDirectory`.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
kirk-marple: In our sample files, the string is ISO 6709 format, such as "+37.7007-121.9416/".  I believe that's what TagGpsLocation already uses?

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
kirk-marple: Nope, those values aren't useful (at least IMO) and could be skipped past.

MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
kirk-marple: Since it was stored in same ISO 6709 format as the GPS found in EXIF data, I'd followed the same pattern.

Doesn't matter to me if you create a new Directory for it or not.

drewnoakes: I kept this in the same directory and added logic to merge the two instances that would otherwise be created.

Issue #231, opened 2019-11-06
Feature Request: HEIF/HEIC/Apple support


billbris:
1: Love the product.  It just worked!  Refreshing.
2: I would like to get at the metadata for my HEIC files from iPhone.
3: Using BitmapMetadata in System.Windows.Media.Imaging I am able to open my .JPEG's and .HEIC's and get at the metadata.  Your package is much more mature and full-featured.  Have you considered the HEIC/Apple problem space?  If so, I'm curious as to why it is not included (am I missing something).  

It just seems that the HEIC/HEIF format will be doing nothing but growing.  

I am curious as to your thoughts.

Also, love the product.

drewnoakes:
@payton implemented HEIF/HEIC support in the Java version of this library a while ago. It shouldn't be too hard to port that implementation to .NET.

https://github.com/drewnoakes/metadata-extractor/pull/314

@billbris would you like to do this?

billbris:
Let me take a peek before I commit (no pun intended).

kastwey:
Hi!

First of all, congratulations on this package, it's wonderful!
Any news about HEIF support? I'm developing a software for my blind wife to read the labels of the photos she has taken with her iPHone (and which is saved in XMP within the HEIC files) with an electron application, but I'm seeing that that part hasn't been ported to .Net yet.

Thanks in advance for the answer!

kastwey:
Hello!

First of all, congratulations on this package, it's wonderful!
Any news about HEIF support? I'm developing a software for my blind wife to read the labels of the photos she has taken with her iPHone (and which is saved in XMP within the HEIC files) with an electron application, but I'm seeing that that part hasn't been ported to .Net yet.

If nobody has started to do it, I could try it :)

Thanks in advance for the answer!

drewnoakes:
@kastwey you're most welcome to port the HEIF/HEIC support from the Java implementation to this library. The contribution would be really appreciated.

Your project sounds really interesting too. I'd like to hear more about it. Will you open source it?

kastwey:
Hi!

Perfect, I already ported all existing heif code (I think) but I've realized that the Exif and XMP support still needs to be implemented, so I'll see if I can add it (although the ISO documentation is crazy).

I'll try to do the PR during the week.

Regarding the project, yes, it will be open source. The initial interface will probably be a disaster (I'm blind too so if something gets out of place or overlaps I'll need help from someone who can see ;) ), but it will be functional :)
[...]

PR #263, opened 2020-07-16
- Added the xmp directory in HEIF format (iso-14496).

kastwey:
- Added the xmp directory in HEIF format (iso-14496). Related to issue #231.
- Fixed a bug where under certain circumstances, the bytes of the exif directory were extended to the next box.
- Fixed an unknown issue when the segments to be parsed start before last position after reading all boxes. Reordered segments bassed on extended offset, not only base offset, so the read will always be in sequential order. This last change fixes the exception thrown in the image file heic/Issue 487.heic.

### DIFF
--- a/MetadataExtractor/Formats/Heif/HeifMetadataReader.cs
+++ b/MetadataExtractor/Formats/Heif/HeifMetadataReader.cs
@@ -48,6 +48,7 @@ namespace MetadataExtractor.Formats.Heif
 
             if (stream.CanSeek)
             {
+                stream.Seek(0, SeekOrigin.Begin);
                 ParseItemSegments();
             }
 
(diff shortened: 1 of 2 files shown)


drewnoakes:
@kastwey thanks very much for your contributions. Are you able to provide sample files that repro these problems?

kastwey:
Hi!

Yes, here it is:
[02. Dins el monestir es veuen 3 de les 8 columnes que el suporten.zip](https://github.com/drewnoakes/metadata-extractor-dotnet/files/4937357/02.Dins.el.monestir.es.veuen.3.de.les.8.columnes.que.el.suporten.zip)

kastwey:
Great! Thanks! :)

PR #264, opened 2020-07-16
Added quick time metadata with no default locale.

kastwey:
Removed the limitation of only default locale is supported. Now, all fields are retrieved, but without locale tag. This should be supported in future. For now, at least, we have the description metadata, which is useful for blind people, I.E., in Apple iPhone, when users labeled the pictures with Voice Over.

### DIFF
--- a/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
+++ b/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
@@ -177,14 +177,8 @@ namespace MetadataExtractor.Formats.QuickTime
                                 a.Reader.Skip(atomSize - 20);
                                 continue;
                             }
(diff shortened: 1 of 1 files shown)


drewnoakes:
@kastwey thanks very much for your contributions. Are you able to provide a sample file that reproduces this problem?

kastwey:
Hi,

Yes, I have a video with this description in quick time directory.

[11. Kastwey kicking a dead man and giving the command to amaterasu_n.zip](https://github.com/drewnoakes/metadata-extractor-dotnet/files/4937284/11.Kastwey.kicking.a.dead.man.and.giving.the.command.to.amaterasu_n.zip)

Thanks!

drewnoakes:
Thanks very much for the PR and the sample file.

kastwey:
Thanks!

PR #266, opened 2020-07-30
Culture sensitivity

### LINKED ISSUE #138: DirectoryExtensions.TryGetDateTime() fails in some cultures
Issue is this function calls [DateTime.TryParseExact()](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tryparseexact?view=netframework-4.7.2#System_DateTime_TryParseExact_System_String_System_String___System_IFormatProvider_System_Globalization_DateTimeStyles_System_DateTime__) with a null format provider. The parse therefore can only succeed if the current culture uses the same separators as the file containing the metadata. From a quick review it looks to me like the intent is to perform invariant parsing, in which case the fix would be to pass CultureInfo.InvariantCulture rather than null.

Workaround is to reimplement the method's code. Personally, I hit this on Arabic so setting CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("ar") on a test thread should provide repro.

patricksadowski:
Closes #138

There are still some string concatenation that do not use an invariant output format. But the main points of the issue are fixed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -222,10 +222,10 @@ dotnet_diagnostic.CA1303.severity = none                          # Literal stri
 # working around bug
 dotnet_diagnostic.RS0041.severity = none                          # Public members should not use oblivious types
 
-# TODO review these diagnostics as many/all are probably valid
-dotnet_diagnostic.CA1304.severity = none                          # Method behaviour depends upon user's locale
-dotnet_diagnostic.CA1305.severity = none                          # Method behaviour depends upon user's locale
-dotnet_diagnostic.CA1307.severity = none                          # Method behaviour depends upon user's locale
+# TODO review these diagnostics as many/all are probably valid, keep current output behavior (english text with localized numbers)
+dotnet_diagnostic.CA1304.severity = none                          # Method behaviour depends upon user's locale (specify CultureInfo f.e. ToLower())
+dotnet_diagnostic.CA1305.severity = none                          # Method behaviour depends upon user's locale (specify IFormatProvider f.e. ToString())
+dotnet_diagnostic.CA1307.severity = none                          # Method behaviour depends upon user's locale (specify StringComparison f.e. StartsWith(), Equals())
 
 # TODO fix these
 dotnet_diagnostic.CA1825.severity = none                          # Avoid unnecessary zero-length array allocations
--- a/MetadataExtractor.Tests/DirectoryTest.cs
+++ b/MetadataExtractor.Tests/DirectoryTest.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Text;
 using MetadataExtractor.Formats.Exif;
 using Xunit;
@@ -103,9 +104,12 @@ namespace MetadataExtractor.Tests
         [InlineData("2002-01-30",                    2002, 1, 30,  0,  0,  0,  0, DateTimeKind.Unspecified)]
         [InlineData("2002-01",                       2002, 1,  1,  0,  0,  0,  0, DateTimeKind.Unspecified)]
         [InlineData("2002",                          2002, 1,  1,  0,  0,  0,  0, DateTimeKind.Unspecified)]
+        [InlineData("2002-01-30T23:59:59.099-08:00", 2002, 1, 31,  7, 59, 59, 99, DateTimeKind.Utc, "ar")]
 #pragma warning restore format
-        public void SetStringAndGetDate(string str, int year, int month, int day, int hour, int minute, int second, int milli, DateTimeKind kind)
+        public void SetStringAndGetDate(string str, int year, int month, int day, int hour, int minute, int second, int milli, DateTimeKind kind, string? culture = null)
         {
+            CultureInfo.CurrentCulture = string.IsNullOrWhiteSpace(culture) ? CultureInfo.CurrentCulture : CultureInfo.GetCultureInfo(culture);
+
             _directory.Set(1, str);
 
             var expected = new DateTime(year, month, day, hour, minute, second, milli, kind);
--- a/MetadataExtractor/DirectoryExtensions.cs
+++ b/MetadataExtractor/DirectoryExtensions.cs
@@ -662,7 +662,7 @@ namespace MetadataExtractor
 
             if (s != null)
             {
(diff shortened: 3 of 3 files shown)


.editorconfig
```diff
@@ -222,11 +222,6 @@ dotnet_diagnostic.CA1303.severity = none                          # Literal stri
 # working around bug
 dotnet_diagnostic.RS0041.severity = none                          # Public members should not use oblivious types
 
-# TODO review these diagnostics as many/all are probably valid
-dotnet_diagnostic.CA1304.severity = none                          # Method behaviour depends upon user's locale
-dotnet_diagnostic.CA1305.severity = none                          # Method behaviour depends upon user's locale
-dotnet_diagnostic.CA1307.severity = none                          # Method behaviour depends upon user's locale
```
drewnoakes, line 228: Is the default "warning" here?

MetadataExtractor.PowerShell/ExtractMetadataCmdlet.cs
```diff
@@ -23,7 +24,7 @@ protected override void ProcessRecord()
         {
             base.ProcessRecord();
 
-            WriteVerbose($"Extracting metadata from file: {FilePath}");
+            WriteVerbose(string.Format(CultureInfo.InvariantCulture, "Extracting metadata from file: {0}", FilePath));
```
drewnoakes, line 27: Were these transformations performed manually or via automation?

drewnoakes, line 27: Pardon my ignorance, but how can culture influence string concatenation? For example, other cases were previously `$"Foo {nameof(Bar)}", which I believe the compiler is within rights to optimise to `"Foo Bar"` at compile time.

drewnoakes:
@patricksadowski thanks for taking the initiative on this. Looks good. Just a few questions to help me understand.

.editorconfig
patricksadowski: I got warnings. The [docs](https://docs.microsoft.com/visualstudio/code-quality/ca1307) categorizes the rules as globalization warnings.

MetadataExtractor.PowerShell/ExtractMetadataCmdlet.cs
patricksadowski: Hand made with ♥. I don't know any tool for that.

patricksadowski: String interpolation uses the current culture. There are issues for cases like
```csharp
var bar = 1.23;
$"Foo {bar}"
```
but (and you are right) not for cases like:
```csharp
var bar = 1.23;
$"Foo {nameof(bar)}"
```
If you like I revert the changes for the cases with `nameof`.

patricksadowski: Finally also cases like
```csharp
var bar = 1.23;
string.Concat("Foo ", bar);
"Foo " + bar;
```
are not taking culture into account.

drewnoakes: I thought perhaps the analyzer came with a codefix.

drewnoakes: Did you only fix warnings produced by the analyzer, or did you search for all string interpolation?

I don't see warnings when reverting some of these changes. Where formattable types are involved, fair enough, but many of these changes involve only strings and the result is a reduction in both readability and performance.

https://sharplab.io/#v2:D4AQjAbABCBMUGECwAoA3qqWYGYaXwAYoBZMACgEsA7AFykoEooBeAPigBIAiAQQCEEUNJQC+UACIBRAGLcA3Jmwg84aOGIlY5DVADOzdlz6Dhe8dLmKU2XPnVhNOHY/2GOPAULTUAhgFsAUwB7ADNyAwtZBSUsWLs1IlIAFio6BnciADoZYIAnf19achNvQii5ABoM61sVeySSAFYXYgNWDg0c/MLi0uFyyWjqg1rlVQJdEghWtw7s3IKikq8Biu5qvyCwiMZGa1EgA

I should point out that I'm also trying to work out whether the analyzer can be improved in order to provide feedback to that team.

patricksadowski: Great idea!

patricksadowski: Visual studio helped me to find all places. I didn't get warnings. Knowing that interpolated strings use the current culture I tried to find them all. Validation was done with `git grep`.

I found another notation the keep the readability.
```csharp
FormattableString.Invariant($"ABC {i} DEF");
```
Do you like that?

drewnoakes:
My original review was focussed on how the analyzer works and whether it could be improved.

Thinking about this some more I realise something deeper. #138 is about a bug in _parsing_ where the input is known to be invariant, but parsing code relies upon the current culture.

This PR fixes that, however it also forces _all formatted output_ to use invariant culture, which is not correct. The library should produce output strings that match the user's culture.

If this was merged as-is, all strings would appear in invariant culture (which is essentially en-US).

patricksadowski:
> This PR fixes that, however it also forces _all formatted output_ to use invariant culture, which is not correct. The library should produce output strings that match the user's culture.

Because of your https://github.com/drewnoakes/metadata-extractor-dotnet/issues/138#issuecomment-624027571 I tried to fix the whole issue. I think I have another unterstanding of culture sensitivity.
If the app outputs English text then also numbers should be formatted like English numbers: `The value equals 1.23` instead of `The value equals 1,23` (German).
If the library supports multiple languages then text shoud be available in each supported language: `The value equals 1.23` _and_ `Der Wert ist gleich 1,23`

To get this done I propose to fix only the main bug of the issue. Your https://github.com/drewnoakes/metadata-extractor-dotnet/issues/138#issuecomment-624027571 should be done in a second PR oder issue. Is it OK for you?

drewnoakes:
> Because of your comment I tried to fix the whole issue

Yeah sorry, I hadn't thought the implications through when I wrote that. It wasn't intended as a strong statement. It may be possible to configure the analyzer to do what we want (via attributes or patterns in the `.editorconfig` file).

> If the app outputs English text then also numbers should be formatted like English numbers

This is debatable. I'm pretty sure if this were changed then we'd see bug reports. Ideally all output strings would be localised.

> To get this done I propose to fix only the main bug of the issue.

That's certainly the safest way forward here. Perhaps we add a bit more context to the TODO in the `.editorconfig` file too.

drewnoakes:
Thanks very much Patrick.

PR #275, opened 2020-11-11
Interpret QuickTime metadata dates

### LINKED ISSUE #257: Support for ISO8601 date time format in iPhone QuickTime files
The QuickTime `.mov` files produced by my iPhone use the `ISO8601` format in the _QuickTime Metadata Header_, e.g. `2019-12-24T22:35:06+0100`.

This format is not supported by the `TryGetDateTime()` extension function (in class `DirectoryExtensions`).

The date can be parsed with this custom date and time format string: `yyyy-MM-ddTHH:mm:sszzz`.

I propose to add this format string to the private list of formats in `string[] _datePatterns` which is used by the `TryGetDateTime()` function.

Thanks a lot for this great library!

patricksadowski:
Closes #257

### DIFF
--- a/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
+++ b/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
@@ -224,6 +224,13 @@ namespace MetadataExtractor.Formats.QuickTime
                                     _ => data
                                 };
 
+                                value = tag switch
+                                {
+                                    QuickTimeMetadataHeaderDirectory.TagCreationDate => DateTime.Parse(((StringValue)value).ToString()),
+                                    QuickTimeMetadataHeaderDirectory.TagLocationDate => DateTime.Parse(((StringValue)value).ToString()),
+                                    _ => value,
+                                };
+
                                 GetMetaHeaderDirectory().Set(tag, value);
                             }
                             else


MetadataExtractor/Formats/QuickTime/QuickTimeMetadataHeaderDescriptor.cs
```diff
@@ -14,6 +14,8 @@ public QuickTimeMetadataHeaderDescriptor(QuickTimeMetadataHeaderDirectory direct
             return tagType switch
             {
                 QuickTimeMetadataHeaderDirectory.TagArtwork => GetArtworkDescription(),
+                QuickTimeMetadataHeaderDirectory.TagCreationDate => Directory.GetDateTime(tagType).ToString("yyyy:MM:dd HH:mm:ss"),
+                QuickTimeMetadataHeaderDirectory.TagLocationDate => Directory.GetDateTime(tagType).ToString("yyyy:MM:dd HH:mm:ss"),
```
drewnoakes, line 18: I think it's better to store `DateTime` in the `QuickTimeMetadataHeaderDirectory` rather than string. This allows the default logic to apply locale-specific formatting for dates.

drewnoakes (commented):
Thanks for fixing this.

drewnoakes:
Thanks!

PR #276, opened 2020-11-14
Added missing Sony model IDs

crener:
Saw that the model ID list for Sony was out of date. Added a load of them using the list from https://exiftool.org/TagNames/Sony.html :)


drewnoakes:
Thanks very much!

PR #291, opened 2021-05-16
Preamble segment handlers

drewnoakes:
Several `IJpegSegmentMetadataReader` implementations behave in a similar manner. This PR pulls out a common base class for them, to reduce duplication and allow consistency.

Also some minor API changes around collection types. These are generally considered internal APIs, though of course some advanced users may have taken a dependency upon these. In such cases, recompiling against the new binary should be enough.


kwhopper:
These are good revisions.

PR #299, opened 2021-06-15
test "IPTC" profile type

### LINKED ISSUE #280: Keyword extraction for PNG files?
I'm trying to pull the keywords from the attached image. If I run Exiftool, the keywords show up correctly (e.g., 'tree' and 'autumn'). But reading it using the .Net Metadata-Extractor doesn't seem to load the keywords; they're not in the `IptcDirectory`, nor anywhere that I can find. I've trawled through the entire structure and tags in the debugger and can't see any keyword tags anywhere. What am I missing? Are PNG keywords supported?

tree

kwhopper:
test if "IPTC" profile is stored as either IPTC IIM or a Photoshop IRB resource

closes #280


drewnoakes (approved):

drewnoakes:
Thanks very much @kwhopper!

PR #314, opened 2022-01-11
Issue#285

### LINKED ISSUE #285: MPEG-2.5 not supported with mp3 files
Hello,

when I tried to ReadMetadata of an mp3 file, a **"MPEG-2.5 not supported"** is returned is there any solution to solve this 
problem ?

jbraendle:
Fixes #285

Hi,
I took a look at the issue and found, that the sample file has ID3V2 tags at the start of the file.
I added support for this form of a mp3 file by skipping the ID3V2 tag before reading the meta data. 
I also changed the MpegAudioTypeChecker class to recognize the ID3V2 tags as mp3 file.
I've shortend the sample file to create a unit test case.

Kind regards 
Jürgen Brändle

### DIFF
--- a/MetadataExtractor/Formats/Mpeg/MpegAudioTypeChecker.cs
+++ b/MetadataExtractor/Formats/Mpeg/MpegAudioTypeChecker.cs
@@ -23,12 +23,20 @@ namespace MetadataExtractor.Formats.Mpeg
                 11 - Layer I
 
             Additional bits contain more information, but are not required for file type identification.
+
+            MP3 with ID3V2-Tags https://id3.org/id3v2.4.0-structure
+            MP3-File with ID3V2-Tagging at start
          */
 
         public int ByteCount => 3;
 
         public Util.FileType CheckType(byte[] bytes)
         {
+
+            // MP3-File with "ID3" at start
+            if (bytes[0] == 0x49 && bytes[1] == 0x44 && bytes[2] == 0x33)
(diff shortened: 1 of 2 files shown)


MetadataExtractor/Formats/Mpeg/MpegAudioTypeChecker.cs
```diff
@@ -23,12 +23,20 @@ internal sealed class MpegAudioTypeChecker : ITypeChecker
...
          */
 
         public int ByteCount => 3;
 
         public Util.FileType CheckType(byte[] bytes)
         {
+
+            // MP3-File with "ID3" at start
+            if (bytes[0] == 0x49 && bytes[1] == 0x44 && bytes[2] == 0x33)
+                return Util.FileType.Mp3;
```
drewnoakes, line 38: I think other types of file can start with ID3 too, though this is an improvement over what we had before.

drewnoakes (approved):
Hi Jürgen,

Thanks very much for this PR.

We avoid adding data files to this source repo, as it can get out of hand very quickly. Instead we use a large corpus of test data, which works well in practice. Please read https://github.com/drewnoakes/metadata-extractor/wiki/Working-with-test-images for more information.

I modified your PR to remove those files, and tweaked the constants and comments a little. I also ran this across the test data and found it fixed a file there.

Thanks again!

PR #317, opened 2022-03-17
Check for exif preamble

### LINKED ISSUE #316: PNG files with exif preamble
Saving a PNG file with exif info in GIMP contains a 'Exif\x0\x0' preamble but this is not handled by the lib.

Example file: https://drive.google.com/file/d/1v8E2g1oiJDFAkWfzuH8u_g4MkEGVjbE6/view?usp=sharing

In PngMetadataReader.cs, adding a check for the exif preamble fixes the issue but I don't know if it is the "correct" way of solving it:
```
else if (keyword == "Raw profile type exif" || keyword == "Raw profile type APP1")
{
    if (TryProcessRawProfile(out _))
    {
        int offset = 0;
        if (ExifReader.StartsWithJpegExifPreamble(textBytes))
            offset = ExifReader.JpegSegmentPreambleLength;

        foreach (var exifDirectory in new ExifReader().Extract(new ByteArrayReader(textBytes, offset)))
            yield return exifDirectory;
    }
    else
    {
        yield return ReadTextDirectory(keyword, textBytes, chunkType);
    }
}
```

reinfallt:
Fixes #316

### DIFF
--- a/MetadataExtractor/Formats/Png/PngMetadataReader.cs
+++ b/MetadataExtractor/Formats/Png/PngMetadataReader.cs
@@ -387,7 +387,11 @@ namespace MetadataExtractor.Formats.Png
                 {
                     if (TryProcessRawProfile(out _))
                     {
-                        foreach (var exifDirectory in new ExifReader().Extract(new ByteArrayReader(textBytes)))
+                        int offset = 0;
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):

drewnoakes:
Perfect, thanks :)

PR #323, opened 2022-06-01
XMP data in png app1 chunk

### LINKED ISSUE #320: PNG file with other keywords/unexpected data
Here is another weird png file: https://drive.google.com/file/d/1mKhXLAafXUb0C6OXrtQtNwRPNVyXlJNn/view?usp=sharing

The parsing mainly has two issues:

1. The keyword for the exif text chunk is "Raw profile type EXIF" instead of "Raw profile type exif". This can be easily fixed by either making the comparison case insensitive or just compare against the other one explicitly.

2. The text chunk with keyword "Raw profile type APP1" actually contains XMP data and not exif data. I did some experiments trying to get it to work (https://github.com/reinfallt/metadata-extractor-dotnet/tree/png_xmp_in_app1_chunk) but I still get an "Error processing XMP data: Unsupported Encoding" from the actual XMP parsing. I didn't debug into the XMP parsing to see what goes wrong.

reinfallt:
This fixes #320 
I had to add an alternative preamble in the XmpReader called JpegSegmentPreamblAlt

### DIFF
--- a/MetadataExtractor/Formats/Xmp/XmpReader.cs
+++ b/MetadataExtractor/Formats/Xmp/XmpReader.cs
@@ -26,9 +26,11 @@ namespace MetadataExtractor.Formats.Xmp
     public sealed class XmpReader : IJpegSegmentMetadataReader
     {
         public const string JpegSegmentPreamble = "http://ns.adobe.com/xap/1.0/\0";
+        public const string JpegSegmentPreamblAlt = "://ns.adobe.com/xap/1.0/\0";
         public const string JpegSegmentPreambleExtension = "http://ns.adobe.com/xmp/extension/\0";
 
         private static byte[] JpegSegmentPreambleBytes { get; } = Encoding.UTF8.GetBytes(JpegSegmentPreamble);
+        private static byte[] JpegSegmentPreambleAltBytes { get; } = Encoding.UTF8.GetBytes(JpegSegmentPreamblAlt);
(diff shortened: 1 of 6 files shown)


MetadataExtractor/Formats/Xmp/XmpReader.cs
```diff
@@ -26,9 +26,11 @@ namespace MetadataExtractor.Formats.Xmp
     public sealed class XmpReader : IJpegSegmentMetadataReader
     {
         public const string JpegSegmentPreamble = "http://ns.adobe.com/xap/1.0/\0";
+        public const string JpegSegmentPreamblAlt = "://ns.adobe.com/xap/1.0/\0";
```
drewnoakes, line 29: Why is the URL scheme (e.g. http or https) omitted here? Is this dealing with a specific type of corruption? If so, why only this particular value? I'm concerned that we are over-fitting to a specific type of corruption here.

reinfallt:
I think you might be correct. There is something weird with this file. I am closing the pull request.

I just spent a couple of hours trying to track down where this file came from. Originally it was part of an old apk file for google plus. It was extracted from an android phone where it had been stored under userdata/data/com.google.android.apps.plus/cache/filter_resources/

Interestingly if I extract the png from the original apk (extracted from the same phone) it doesn't have any of the metadata sections that the file from the cache on the phone had.

PR #324, opened 2022-06-20
GeoTIFF fixes + add test

don-vip:
Ported bug fixes and tests from https://github.com/drewnoakes/metadata-extractor/pull/586

I got a behaviour difference between Java and .NET regarding the description of TagGeoTiffGeoKeys, I had to comment this check to get the test to pass, I don't understand why:

```csharp
    //Assert.Equal("[32 values]", exif.GetDescription(ExifDirectoryBase.TagGeoTiffGeoKeys));
```


drewnoakes:
This looks great with one caveat: we don't add test images to this repo. Instead, we add them to https://github.com/drewnoakes/metadata-extractor-images, where the regression tests track changes over time from each of the Java and .NET implementations, along with a diff of the outputs from each. Ideally they should produce the same values.

Adding images to this repo (or the Java repo) increase the size significantly.

drewnoakes:
I moved the code changes here to #374. Thanks again!

don-vip:
Thank you!

PR #327, opened 2023-01-24
Remove excessive bounds checking in IndexedReader

### LINKED ISSUE #62: Excessive bounds checking when reading multi-byte values

drewnoakes:
Fixes #62

Adds a new abstract overload for reading a single byte. That method will not do any explicit bounds validation, so callers must do that beforehand. In cases where multiple bytes are read, the validation can be performed once per value, rather than once per byte.

### DIFF
--- a/MetadataExtractor/IO/IndexedCapturingReader.cs
+++ b/MetadataExtractor/IO/IndexedCapturingReader.cs
@@ -131,12 +131,21 @@ namespace MetadataExtractor.IO
 
         public override int ToUnshiftedOffset(int localOffset) => localOffset;
 
-        public override byte GetByte(int index)
+        private void GetPosition(int index, out int chunkIndex, out int innerIndex)
         {
-            ValidateIndex(index, 1);
+#if NET35 || NET45 || NETSTANDARD2_0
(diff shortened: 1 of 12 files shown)


drewnoakes:
This has some failures in the (manual) regression tests that need to be investigated.

MetadataExtractor/IO/IndexedCapturingReader.cs
```diff
@@ -151,8 +160,7 @@ public override byte[] GetBytes(int index, int count)
             var toIndex = 0;
             while (remaining != 0)
             {
-                var fromChunkIndex = fromIndex / _chunkLength;
-                var fromInnerIndex = fromIndex % _chunkLength;
+                GetPosition(index, out int fromChunkIndex, out int fromInnerIndex);
```
drewnoakes, line 163: This is the problem:

```suggestion
                GetPosition(fromIndex, out int fromChunkIndex, out int fromInnerIndex);
```

PR #329, opened 2023-04-29
Fixes Issue #328

### LINKED ISSUE #328: Reading UTF-8 from iTXt PNGChunk
In PNGMetadataReader.cs mentions:

        /// Note that "iTXt" chunks use UTF-8 encoding (https://www.w3.org/TR/PNG/#11iTXt).

And the document at that URL states:

> The translated keyword and text both use the UTF-8 encoding

However at L:249, the keyword is read using _latin1Encoding, and the text is read in ReadTextDirectory using _latin1Encoding.

This leads to UTF-8 characters such as emoji being decoded improperly.

RupertAvery:
Fixes #328

Parse the keyword and text directory using utf8 encoding if PNG Chunk Type is iTXt

### DIFF
--- a/MetadataExtractor/Formats/Png/PngMetadataReader.cs
+++ b/MetadataExtractor/Formats/Png/PngMetadataReader.cs
@@ -249,7 +249,8 @@ namespace MetadataExtractor.Formats.Png
             else if (chunkType == PngChunkType.iTXt)
             {
                 var reader = new SequentialByteArrayReader(bytes);
-                var keyword = reader.GetNullTerminatedStringValue(maxLengthBytes: 79).ToString(_utf8Encoding);
+                var keywordStringValue = reader.GetNullTerminatedStringValue(maxLengthBytes: 79);
+                var keyword = keywordStringValue.ToString(_utf8Encoding);
                 var compressionFlag = reader.GetSByte();
                 var compressionMethod = reader.GetSByte();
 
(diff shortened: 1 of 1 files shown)


MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -90,6 +90,7 @@ public static DirectoryList ReadMetadata(Stream stream)
         /// For more guidance: http://www.w3.org/TR/PNG-Decoders.html#D.Text-chunk-processing
         /// </summary>
         private static readonly Encoding _latin1Encoding = Encoding.GetEncoding("ISO-8859-1");
+        private static readonly Encoding _utf8Encoding = Encoding.GetEncoding("UTF-8");
```
drewnoakes, line 93: I think we could use `Encoding.UTF8`, and inline the field.

drewnoakes (commented):
Thanks, this looks great. I'll try it on the regression test data set before merging.

MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -248,7 +249,7 @@ private static IEnumerable<Directory> ProcessChunk(PngChunk chunk)
             else if (chunkType == PngChunkType.iTXt)
             {
                 var reader = new SequentialByteArrayReader(bytes);
-                var keyword = reader.GetNullTerminatedStringValue(maxLengthBytes: 79).ToString(_latin1Encoding);
+                var keyword = reader.GetNullTerminatedStringValue(maxLengthBytes: 79).ToString(_utf8Encoding);
```
drewnoakes, line 252: There's actually now an issue slightly below here. The `bytesLeft` value was based on the length of the string in bytes, which for latin1 is the same as the length of the string in characters. With UTF-8 that's not the case. I'll patch this up and push to your PR.

drewnoakes:
I ran this against the regression test data set and it now successfully parses a bunch of previously broken values. I see no downside to this anywhere. Great stuff, thanks!

drewnoakes (approved):

PR #349, opened 2023-11-22
Fix NikonPictureControl1Directory parsing

drewnoakes:
Fix NikonPictureControl1Directory string reading. These fixed-width, null-terminated strings. We don't want those nulls in the string we return.

Also some minor tidying and adding a unit test.

### DIFF
--- a/MetadataExtractor.Tests/.editorconfig
+++ b/MetadataExtractor.Tests/.editorconfig
@@ -1,0 +1,3 @@
+[*.cs]
+
+dotnet_diagnostic.CA1707.severity = none                    # Identifiers should not contain underscores
--- a/MetadataExtractor.Tests/MetadataExtractor.Tests.csproj
+++ b/MetadataExtractor.Tests/MetadataExtractor.Tests.csproj
@@ -36,7 +36,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Using Include="Xunit"/>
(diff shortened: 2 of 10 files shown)


gtiwari333:
@drewnoakes  i am not sure how .NET works. but the code looks identical in Java and .NET. 

Based on what i understand, this change will not read the value correctly. It will ONLY work if there are no null-terminated string in metadata.

```
        directory.Set(TagPictureControlVersion, reader.GetNullTerminatedStringValue(4));
        directory.Set(TagPictureControlName, reader.GetNullTerminatedStringValue(20));
        directory.Set(TagPictureControlBase, reader.GetNullTerminatedStringValue(20));
```

The issue is similar to what i described on https://github.com/drewnoakes/metadata-extractor/pull/636#issuecomment-1822998203.

In order to fix it, we need to either check the length of returned string and move the cursor after each read. Or, use a code similar to `getNullTerminatedStringAndSkipToNextPosition`.

drewnoakes:
You're right, thanks. Fixed in https://github.com/drewnoakes/metadata-extractor-dotnet/pull/350 specifically e7cbf06e88efb7f6fca9dd83dd1214d23aef2329 and bff31ad2a582d6bd83a84c2f960ce30234f8cdad. I'll review the Java version, run the regression tests and get a fix up and another release out.

PR #353, opened 2023-12-12
Use file scoped namespaces

iamcarbon:
@drewnoakes Ready for feedback / review. This PR contains no functional changes.

File scoped namespaces improve the codebase readability, particularly on smaller screens (like laptops).


iamcarbon:
Here's a direct link to review, ignoring whitespace changes:

https://github.com/drewnoakes/metadata-extractor-dotnet/pull/353/files?diff=unified&w=1

drewnoakes:
Thanks but I'm not inclined to change this. Such broad changes can interfere with trawling through history to understand why code exists as is.. While I like the language feature, I'd prefer to only use it for new code.

PR #361, opened 2024-01-18
Spanify various things and use UTF-8 literals

iamcarbon:
Posting for early feedback. 

There is one public API change, which I'm happy to revert if we have a no-breaking changes policy.

`protected byte[] PreambleBytes` was changed to `protected ReadOnlySpan<byte> PreambleBytes`

This also prevents the mutation of the underlying data.


iamcarbon:
@drewnoakes Ready for review.

drewnoakes:
Thanks @iamcarbon this is great. I've started spanifying a few things locally too, though we don't have much overlap which is great. I'm looking mostly at the deserialization stuff, using `BinaryPrimitives` rather than hand-rolled functions. The built-in ones will be faster.

> There is one public API change

I think that's fine, as it's not an API I'd expect many to use directly (if any), though I'm surprised it didn't require a change to the API files.

drewnoakes (approved):

PR #363, opened 2024-01-19
Spanify (part 2)

iamcarbon:
This PR updates a few more preambles with statically initialized data, and updates the uses of ByteArrayUtil to use built-in vectorized span functions. 

[Breaking] This PR also removes the now unused ByteArrayUtil, which was previously public.


iamcarbon:
@drewnoakes Ready for review / feedback.

drewnoakes (approved):
Looks great!

PR #366, opened 2024-01-20
Use BinaryPrimitives

iamcarbon:
Ready for review.


MetadataExtractor.Tests/DirectoryTest.cs
```diff
@@ -49,8 +49,8 @@ public void UnderlyingInt()
             Assert.Equal(value.ToString(), _directory.GetString(tagType));
             Assert.True(_directory.TryGetRational(tagType, out Rational rational));
             Assert.Equal(new Rational(value, 1), rational);
-            Assert.Equal([value], _directory.GetInt32Array(tagType));
-            Assert.Equal(new[] { unchecked((byte)value) }, _directory.GetByteArray(tagType));
+            Assert.Equal((int[])[value], _directory.GetInt32Array(tagType));
```
drewnoakes, line 52: I'm curious. Did Visual Studio recommend the original conversation to a collection literal? If so I'll file a big against the analyser for suggesting code that changes semantics.

drewnoakes: I'm on my phone so haven't investigated, but can we write this instead:


```
Assert.Equal<int>([1, 2], foo);
```

If not then maybe the original form would be more common.

```
new int[] { 1, 2 }
```

drewnoakes (commented):
Looks good. I wonder if we could reformat one pattern around collection expressions. Can merge once we resolve that question.

I'm spanifying the reader classes at the moment so please hold off on touching those in any significant way.

Thanks for the great PRs. They add clear value and are the perfect size for quick review.

MetadataExtractor.Tests/DirectoryTest.cs
iamcarbon: Yep, it was recommended by VS. I agree that we may should update the analyzer rules here. 


> I'm curious. Did Visual Studio recommend the original conversation to a collection literal? If so I'll file a big against the analyser for suggesting code that changes semantics.

Yep, it was recommended by VS. I agree that we may should update the analyzer rules here.

drewnoakes (approved):

PR #367, opened 2024-01-21
Spanify (3/n)

iamcarbon:
- Updates mutable byte[] data in HuffmanTablesDirectory with statistically initialized data
- Adds and utilizes JpegSegment.Span property to reduce need for AsSpan() 
- Uses static JpegSegmentPreambleBytes in DuckyReader


iamcarbon:
@drewnoakes Ready for review / feedback. CI is failing due to an ambiguous xunit method call (resolved in #366)

MetadataExtractor/Formats/Photoshop/DuckyReader.cs
```diff
@@ -9,14 +9,15 @@ namespace MetadataExtractor.Formats.Photoshop
     public sealed class DuckyReader : IJpegSegmentMetadataReader
     {
         public const string JpegSegmentPreamble = "Ducky";
```
drewnoakes, line 11: I wonder if we still need these preamble strings. Can be a future PR.

MetadataExtractor/Formats/Photoshop/DuckyReader.cs
```diff
@@ -9,14 +9,15 @@ namespace MetadataExtractor.Formats.Photoshop
...
+        internal static ReadOnlySpan<byte> JpegSegmentPreambleBytes => "Ducky"u8;
 
-        ICollection<JpegSegmentType> IJpegSegmentMetadataReader.SegmentTypes { get; } = new[] { JpegSegmentType.AppC };
+        ICollection<JpegSegmentType> IJpegSegmentMetadataReader.SegmentTypes { get; } = [JpegSegmentType.AppC];
```
drewnoakes, line 14: We should bump these to IReadOnlyCollection. Can be a future PR.

MetadataExtractor/Formats/Exif/ExifReader.cs
```diff
@@ -17,7 +17,7 @@ public sealed class ExifReader : JpegSegmentWithPreambleMetadataReader
 
         private static ReadOnlySpan<byte> ExifPreamble => "Exif\x0\x0"u8;
 
-        public static bool StartsWithJpegExifPreamble(byte[] bytes) => bytes.AsSpan().StartsWith(ExifPreamble);
+        public static bool StartsWithJpegExifPreamble(byte[] bytes) => bytes.AsSpan.StartsWith(ExifPreamble);
```
drewnoakes, line 20: Should AsSpan -> Span?

drewnoakes (commented):
The other PR has merged so it should be possible to get the checks here to pass now.

Left some comments for potential future work too. We can file issues to track them if we feel they're valid.

MetadataExtractor/Formats/Photoshop/DuckyReader.cs
iamcarbon: If we're OK with a breaking change, it would be great to replace them all with the UTF8 encoded version.

MetadataExtractor/Formats/Photoshop/DuckyReader.cs
iamcarbon: Annoyingly, HashSet<JpegSegmentType> doesn't implement IReadOnlyCollection, which we use in JpegReader.

MetadataExtractor/Formats/Exif/ExifReader.cs
iamcarbon: UGH! Dumb mistake! And good catch. Fixed.

MetadataExtractor/Formats/Photoshop/DuckyReader.cs
drewnoakes: Looks like they're used in the powershell code. That could be revisited in future too.

MetadataExtractor/Formats/Photoshop/DuckyReader.cs
drewnoakes: Looking at usages of the property, they're all enumerations. So `IReadOnlyCollection` or `IEnumerable` are both candidates, and `JpegReader` would be changed to no longer use `HashSet<T>`.

drewnoakes (approved):

PR #376, opened 2024-01-22
Update JpegSegmentPreamble to UTF-8 spans

iamcarbon:
@drewnoakes Ready for review / feedback.

Note: The AdobeJpegReader compares the Preamble using a case-insensitive comparer in ReadJpegSegments, while it uses an ordinal comparer when reading segments. From what I can tell in the specifications, this should be exact, but I didn't want to change behavior here.

### DIFF
--- a/MetadataExtractor.PowerShell/MetadataExtractor.PowerShell.csproj
+++ b/MetadataExtractor.PowerShell/MetadataExtractor.PowerShell.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net45</TargetFramework>
+    <TargetFramework>net48</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/MetadataExtractor.PowerShell/MetadataExtractor.PowerShell.csproj
+++ b/MetadataExtractor.PowerShell/MetadataExtractor.PowerShell.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net48</TargetFramework>
+    <TargetFramework>net462</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/MetadataExtractor/Formats/Exif/ExifReader.cs
+++ b/MetadataExtractor/Formats/Exif/ExifReader.cs
@@ -13,16 +13,14 @@ namespace MetadataExtractor.Formats.Exif
     /// <author>Drew Noakes https://drewnoakes.com</author>
     public sealed class ExifReader : JpegSegmentWithPreambleMetadataReader
     {
-        public const string JpegSegmentPreamble = "Exif\x0\x0";
+        public static ReadOnlySpan<byte> JpegSegmentPreamble => "Exif\x0\x0"u8;
 
-        private static ReadOnlySpan<byte> ExifPreamble => "Exif\x0\x0"u8;
+        public static bool StartsWithJpegExifPreamble(byte[] bytes) => bytes.AsSpan().StartsWith(JpegSegmentPreamble);
 
-        public static bool StartsWithJpegExifPreamble(byte[] bytes) => bytes.AsSpan().StartsWith(ExifPreamble);
-
-        public static int JpegSegmentPreambleLength => ExifPreamble.Length;
+        public static int JpegSegmentPreambleLength => JpegSegmentPreamble.Length;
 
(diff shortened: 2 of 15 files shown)


MetadataExtractor.PowerShell/MetadataExtractor.PowerShell.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net45</TargetFramework>
+    <TargetFramework>net48</TargetFramework>
```
drewnoakes, line 4: What necessitated this?

MetadataExtractor/Formats/Exif/ExifReader.cs
```diff
@@ -13,16 +13,14 @@ namespace MetadataExtractor.Formats.Exif
     /// <author>Drew Noakes https://drewnoakes.com</author>
     public sealed class ExifReader : JpegSegmentWithPreambleMetadataReader
     {
-        public const string JpegSegmentPreamble = "Exif\x0\x0";
+        public static ReadOnlySpan<byte> JpegSegmentPreamble => "Exif\x0\x0"u8;
```
drewnoakes, line 16: If we only targeted .NET 7+, this member could be on the interface 😀

MetadataExtractor/Util/ByteTrie.cs
```diff
@@ -90,7 +90,7 @@ public void Add(T value, params byte[][] parts)
         }
 
         /// <summary>Store the given value at the specified path.</summary>
-        internal void Add(T value, ReadOnlySpan<byte> part)
+        public void Add(T value, ReadOnlySpan<byte> part)
```
drewnoakes, line 93: Is the byte[] overload still used?

Does this need to be added to the public API files?

iamcarbon: This was a brand new member to avoid allocation, the params byte[] array overload still used.

MetadataExtractor/Formats/Xmp/XmpReader.cs
```diff
@@ -19,11 +19,8 @@ namespace MetadataExtractor.Formats.Xmp
     /// <author>Drew Noakes https://drewnoakes.com</author>
     public sealed class XmpReader : IJpegSegmentMetadataReader
     {
-        public const string JpegSegmentPreamble = "http://ns.adobe.com/xap/1.0/\0";
-        public const string JpegSegmentPreambleExtension = "http://ns.adobe.com/xmp/extension/\0";
-
-        private static ReadOnlySpan<byte> JpegSegmentPreambleBytes => "http://ns.adobe.com/xap/1.0/\0"u8;
-        private static ReadOnlySpan<byte> JpegSegmentPreambleExtensionBytes => "http://ns.adobe.com/xmp/extension/\0"u8;
+        public static ReadOnlySpan<byte> JpegSegmentPreamble=> "http://ns.adobe.com/xap/1.0/\0"u8;
```
drewnoakes, line 22:
```suggestion
        public static ReadOnlySpan<byte> JpegSegmentPreamble => "http://ns.adobe.com/xap/1.0/\0"u8;
```

MetadataExtractor/Util/ByteTrie.cs
drewnoakes: Is there a diagnostic here about the public member not being in the API files?

drewnoakes (commented):
The Adobe case is a shame. It's the only one that doesn't fit. What do you think about changing the interface to include a method that validates whether a given segments bytes are a match or not? Then the Adobe one could implement it's policy within that method. The dictionary in the PowerShell code would be swapped for iteration through known extractors, testing for a match.

MetadataExtractor/Util/ByteTrie.cs
iamcarbon: Annoying, this analyzer randomly stops reporting for me in Visual Studio, and I have to catch these manually. Updated!

MetadataExtractor/Formats/Xmp/XmpReader.cs
iamcarbon: Fixed

MetadataExtractor/Formats/Exif/ExifReader.cs
drewnoakes: I work a lot in the VS codebase which still targets .NET Framework. I really enjoy when I can use the newer language features.

MetadataExtractor/Util/ByteTrie.cs
drewnoakes: Weird. There should at least be a build warning, and CI should fail. I'll take a look some time.

drewnoakes (approved):
Awesome, thanks!

PR #380, opened 2024-01-27
Spanify indexed reader classes

drewnoakes:
Allows avoidance of temporary byte arrays, reduces the number of virtual calls to `GetByteInternal`, and uses conversion methods optimised by the framework (e.g. `BinaryPrimitives`).

### DIFF
--- a/MetadataExtractor/IO/ByteArrayReader.cs
+++ b/MetadataExtractor/IO/ByteArrayReader.cs
@@ -33,9 +33,11 @@ namespace MetadataExtractor.IO
 
         public override long Length => _buffer.Length - _baseOffset;
 
-        protected override byte GetByteInternal(int index)
+        public override void GetBytes(int index, Span<byte> bytes)
         {
-            return _buffer[index + _baseOffset];
+            ValidateIndex(index, bytes.Length);
+
+            _buffer.AsSpan().Slice(index + _baseOffset, bytes.Length).CopyTo(bytes);
         }
 
         protected override void ValidateIndex(int index, int bytesRequested)
@@ -44,21 +46,12 @@ namespace MetadataExtractor.IO
                 throw new BufferBoundsException(ToUnshiftedOffset(index), bytesRequested, _buffer.Length);
         }
 
-        protected override bool IsValidIndex(int index, int bytesRequested)
+        private bool IsValidIndex(int index, int bytesRequested)
         {
             return
                 bytesRequested >= 0 &&
                 index >= 0 &&
                 index + (long)bytesRequested - 1L < Length;
-        }
-
-        public override byte[] GetBytes(int index, int count)
-        {
-            ValidateIndex(index, count);
-
-            var bytes = new byte[count];
-            Array.Copy(_buffer, index + _baseOffset, bytes, 0, count);
-            return bytes;
         }
     }
 }
--- a/MetadataExtractor/IO/IndexedReader.cs
+++ b/MetadataExtractor/IO/IndexedReader.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
 
-// TODO always read bytes in order which may assist memory read patterns
+using System.Buffers.Binary;
 
 namespace MetadataExtractor.IO
 {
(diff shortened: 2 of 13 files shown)


drewnoakes:
@iamcarbon please take a look if you have a chance.

MetadataExtractor/IO/IndexedReader.cs
```diff
@@ -307,33 +255,77 @@ public ulong GetUInt64(int index)
...
+#if NET45 || NETSTANDARD1_3
             var bytes = GetBytes(index, bytesRequested);
+
             return encoding.GetString(bytes, 0, bytes.Length);
+#else
+            Span<byte> bytes = bytesRequested < 256 ? stackalloc byte[bytesRequested] : new byte[bytesRequested];
+
+            GetBytes(index, bytes);
+
+            return encoding.GetString(bytes);
```
iamcarbon, line 327: If we bump netframework to 462, we may consider adding a polyfill for Encoding.GetString to have two paths here. This would remain allocation free, and simplify ^.

For example:

```csharp
#if NETFRAMEWORK || NETSTANDARD1_3
internal static class EncodingExtensions
{
    public unsafe static string GetString(this Encoding encoding, ReadOnlySpan<byte> bytes)
    {
        fixed (byte* pBytes = bytes)
        {
            return encoding.GetString(pBytes, bytes.Length);
        }
    }
}
#endif
```

MetadataExtractor/IO/IndexedReader.cs
```diff
@@ -271,30 +236,13 @@ public long GetInt64(int index)
...
-                (ulong)GetByteInternal(index + 2) << 16 |
-                (ulong)GetByteInternal(index + 1) <<  8 |
-                      GetByteInternal(index    );
+            Span<byte> bytes = stackalloc byte[8];
+
+            GetBytes(index, bytes);
+
+            return IsMotorolaByteOrder
+                ? BinaryPrimitives.ReadUInt64BigEndian(bytes)
+                : BinaryPrimitives.ReadUInt64LittleEndian(bytes);
```
iamcarbon, line 245: Much nicer!

MetadataExtractor/IO/ByteArrayReader.cs
```diff
@@ -44,21 +46,12 @@ protected override void ValidateIndex(int index, int bytesRequested)
...
         }
-
-        public override byte[] GetBytes(int index, int count)
-        {
-            ValidateIndex(index, count);
-
-            var bytes = new byte[count];
-            Array.Copy(_buffer, index + _baseOffset, bytes, 0, count);
-            return bytes;
-        }
```
iamcarbon, line 62: Good to see the allocating functions killed.

iamcarbon (approved):

iamcarbon:
Looks great. Just one comment about a future simplification.

MetadataExtractor/IO/IndexedReader.cs
drewnoakes: That's a great approach and we could use it in a few places.

I'll have a think about bumping the target further. Are there other APIs in 462 you know of that's be good to use?

iamcarbon: Array.Empty<T> is the only other one I've missed so far. That would allow us to kill Util.Empty. It's possible that the lowering of [] may also be improved on net462.

drewnoakes: I'm good with bumping to `net462`. 4.5.2 went out of support in 2022. 4.6.2 is supported until 2027. I'll file a different PR for that then rebase and update this one.

drewnoakes:
I'll investigate the encoding changes in a separate PR.

drewnoakes:
@iamcarbon let's get 2.9.0 out soon. If there are any more public API changes you think we'd want, specifically related to performance in response to targeting newer frameworks, let me know.

I'm thinking of getting a prerelease package out today. We can also test the new infra for publishing to NuGet by creating releases here on GitHub.

MetadataExtractor/IO/IndexedReader.cs
drewnoakes: https://github.com/drewnoakes/metadata-extractor-dotnet/pull/386

drewnoakes:
Pre-release package is available, per https://github.com/drewnoakes/metadata-extractor-dotnet/pull/389#issuecomment-1913993882

PR #389, opened 2024-01-29
Remove explicit Version MSBuild property

drewnoakes:
Now that we're moving to building and pushing packages via GitHub actions, we will take the version from the release itself. This means we don't want it to be specified in the file directly.

I like this change as it means we don't have to keep the props file in sync with the current version, simplifying the release process somewhat. We only have to bump the version in the README which directs users to install the current version. That can always be done after release if missed.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,6 @@
 <Project>
 
   <PropertyGroup>
-    <VersionPrefix>2.8.1</VersionPrefix>
     <DebugType>portable</DebugType>
     <LangVersion>preview</LangVersion>
     <Nullable>enable</Nullable>


drewnoakes:
> Now that we're moving to building and pushing packages via GitHub actions

The first run of this just completed successfully.

- Creating release https://github.com/drewnoakes/metadata-extractor-dotnet/releases/tag/2.9.0-rc1
- Triggered run https://github.com/drewnoakes/metadata-extractor-dotnet/actions/runs/7691635459/job/20957241341
- Publishing package https://www.nuget.org/packages/MetadataExtractor/2.9.0-rc1

This is a huge improvement over the previous manual release process.

Every 12 months I'll need to refresh the NuGet API key in this repo's secrets.

drewnoakes:
Validated package contents and metadata, including SourceLink and determinism:

image

drewnoakes:
@iamcarbon once you're happy with the changes for 2.9.0 (especially API-changes) we can push out a release.

PR #390, opened 2024-01-29
Improve Code Quality / Tiding

iamcarbon:
- Simplifies string.Join calls (eliminates various array allocations)
- Uses new GetString polyfill
- Removes Empty class (using empty collection expression instead)
- Eliminates various byte allocation
- Updates ShouldAcceptList to use utf-8 bytes
- Updates StartsWithJpegPreamble to accept ReadOnlySpan<byte>
- Switches on integer values to avoid a string allocation

### DIFF
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
@@ -651,15 +651,15 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             if (Directory.GetObject(OlympusCameraSettingsMakernoteDirectory.TagGradation) is not short[] values || values.Length < 3)
                 return null;
 
-            var join = $"{values[0]} {values[1]} {values[2]}";
-            var ret = join switch
+            var ret = (values[0], values[1], values[3]) switch
             {
-                "0 0 0" => "n/a",
-                "-1 -1 1" => "Low Key",
-                "0 -1 1" => "Normal",
-                "1 -1 1" => "High Key",
-                _ => "Unknown (" + join + ")",
+                (0, 0, 0) => "n/a",
+                (-1, -1, 1) => "Low Key",
+                (0, -1, 1) => "Normal",
+                (1, -1, 1) => "High Key",
+                _ => $"Unknown ({values[0]} {values[1]} {values[2]})"
             };
+
             if (values.Length > 3)
             {
                 if (values[3] == 0)
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
@@ -77,13 +77,11 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             if (Directory.GetObject(OlympusFocusInfoMakernoteDirectory.TagExternalFlash) is not ushort[] values || values.Length < 2)
                 return null;
 
-            var join = $"{values[0]} {values[1]}";
-
-            return join switch
+            return (values[0], values[1]) switch
             {
-                "0 0" => "Off",
-                "1 0" => "On",
-                _ => "Unknown (" + join + ")",
+                (0, 0) => "Off",
+                (1, 0) => "On",
+                _ => $"Unknown ({values[0]} {values[1]})"
             };
         }
 
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
@@ -30,7 +30,7 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
                 OlympusFocusInfoMakernoteDirectory.TagMacroLed => GetMacroLedDescription(),
                 OlympusFocusInfoMakernoteDirectory.TagSensorTemperature => GetSensorTemperatureDescription(),
                 OlympusFocusInfoMakernoteDirectory.TagImageStabilization => GetImageStabilizationDescription(),
-                _ => base.GetDescription(tagType),
+                _ => base.GetDescription(tagType)
             };
         }
 
@@ -107,15 +107,13 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             if (values.Length == 0)
                 return null;
 
-            var join = $"{values[0]}" + (values.Length > 1 ? $"{values[1]}" : "");
-
-            return join switch
+            return (values[0], values.Length > 1 ? values[1] : -1) switch
             {
-                "0" => "Off",
-                "1" => "On",
-                "0 0" => "Off",
-                "1 0" => "On",
-                _ => "Unknown (" + join + ")",
+                (0, -1) => "Off",
+                (1, -1) => "On",
+                (0, 0) => "Off",
+                (1, 0) => "On",
+                _ => $"Unknown ({string.Join(" ", values)})"
             };
         }
 
--- a/MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
+++ b/MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
@@ -109,7 +109,7 @@ namespace MetadataExtractor.Formats.Exif.Makernotes
             if (values.Length == 0)
                 return null;
 
-            var join = $"{values[0]}" + (values.Length > 1 ? $"{ values[1]}" : "");
+            var join = $"{values[0]}" + (values.Length > 1 ? $"{values[1]}" : "");
 
             return join switch
             {
--- a/MetadataExtractor/Formats/Heif/HeicImagePropertyDescriptor.cs
+++ b/MetadataExtractor/Formats/Heif/HeicImagePropertyDescriptor.cs
@@ -15,7 +15,7 @@ namespace MetadataExtractor.Formats.Heif
                     return Directory.GetString(tagType) + " degrees";
                 case HeicImagePropertiesDirectory.TagPixelDepths:
                     var o = Directory.GetObject(HeicImagePropertiesDirectory.TagPixelDepths);
-                    return o is null ? null : string.Join(" ", ((byte[])o).Select(i => i.ToString()).ToArray());
+                    return o is null ? null : string.Join(" ", ((byte[])o).Select(i => i.ToString()));
                 case HeicImagePropertiesDirectory.TagColorFormat:
                     return TypeStringConverter.ToTypeString(Directory.GetUInt32(HeicImagePropertiesDirectory.TagColorFormat));
                 case HeicImagePropertiesDirectory.TagColorPrimaries:
--- a/MetadataExtractor/Formats/Heif/HeicImagePropertyDescriptor.cs
+++ b/MetadataExtractor/Formats/Heif/HeicImagePropertyDescriptor.cs
@@ -15,7 +15,7 @@ namespace MetadataExtractor.Formats.Heif
                     return Directory.GetString(tagType) + " degrees";
                 case HeicImagePropertiesDirectory.TagPixelDepths:
                     var o = Directory.GetObject(HeicImagePropertiesDirectory.TagPixelDepths);
-                    return o is null ? null : string.Join(" ", ((byte[])o).Select(i => i.ToString()));
+                    return o is byte[] bytes ? string.Join(" ", bytes) : null;
                 case HeicImagePropertiesDirectory.TagColorFormat:
                     return TypeStringConverter.ToTypeString(Directory.GetUInt32(HeicImagePropertiesDirectory.TagColorFormat));
                 case HeicImagePropertiesDirectory.TagColorPrimaries:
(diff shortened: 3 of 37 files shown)


iamcarbon:
@drewnoakes Ready for review. This is the last of the changes proposed for 2.9.

drewnoakes (commented):
Looking good. Thanks for fixing all these. Left some comments. See what you think and we'll get this merged.

MetadataExtractor/Formats/Heif/HeicImagePropertyDescriptor.cs
```diff
@@ -15,7 +15,7 @@ public class HeicImagePropertyDescriptor(HeicImagePropertiesDirectory directory)
                     return Directory.GetString(tagType) + " degrees";
                 case HeicImagePropertiesDirectory.TagPixelDepths:
                     var o = Directory.GetObject(HeicImagePropertiesDirectory.TagPixelDepths);
-                    return o is null ? null : string.Join(" ", ((byte[])o).Select(i => i.ToString()).ToArray());
+                    return o is null ? null : string.Join(" ", ((byte[])o).Select(i => i.ToString()));
```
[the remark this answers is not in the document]
iamcarbon: Yep! updated.

MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
```diff
@@ -109,7 +107,7 @@ public sealed class OlympusFocusInfoMakernoteDescriptor(OlympusFocusInfoMakernot
             if (values.Length == 0)
                 return null;
 
-            var join = $"{values[0]}" + (values.Length > 1 ? $"{ values[1]}" : "");
+            var join = $"{values[0]}" + (values.Length > 1 ? $"{values[1]}" : "");
```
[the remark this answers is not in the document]
iamcarbon: Yep, looks like an existing bug. I had originally avoid changing the behavior on these, and just fixed the formatting error. I updated this to use a switch statement. 

NOTE: the values array ushort[], so -1 can be safely used as no-value placeholder.

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
```diff
@@ -9,12 +9,11 @@ public static class QuickTimeReaderExtensions
...
-            sb.Append((char)reader.GetByte());
-            sb.Append((char)reader.GetByte());
-            sb.Append((char)reader.GetByte());
-            sb.Append((char)reader.GetByte());
-            return sb.ToString();
+            Span<byte> bytes = stackalloc byte[4];
+
+            reader.GetBytes(bytes);
+
+            return new string([(char)bytes[0], (char)bytes[1], (char)bytes[2], (char)bytes[3]]);
```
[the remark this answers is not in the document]
iamcarbon: .NETStandard2.1 has a System.ReadOnlySpan<char> overload that stack allocates. However, agree on updating to ASCII.GetString to avoid the char[] allocation on other targets.

The callers here expect this function to produce a string, so not much else to optimize here.

Screenshot 2024-01-29 at 6 55 06 PM

MetadataExtractor/Formats/Riff/RiffReader.cs
```diff
@@ -70,7 +70,7 @@ private static void ProcessChunks(SequentialReader reader, long maxPosition, IRi
                 {
                     if (chunkSize < 4)
                         break;
-                    string listName = reader.GetString(4, Encoding.ASCII);
+                    ReadOnlySpan<byte> listName = reader.GetBytes(4);
```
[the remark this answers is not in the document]
iamcarbon: I avoided a stackalloc in this case, since we're in a while statement. but let me refactor this to eliminate this safely.

Screenshot 2024-01-29 at 7 01 00 PM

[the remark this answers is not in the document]
iamcarbon: Updated!

MetadataExtractor/Formats/Iptc/Iso2022Converter.cs
```diff
@@ -91,7 +91,7 @@ public static class Iso2022Converter
             {
                 try
                 {
-                    var s = encoding.GetString(bytes, 0, bytes.Length);
+                    var s = encoding.GetString(bytes);
```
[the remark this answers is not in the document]
iamcarbon: For now, I updated this to use the ArrayPool to reduce GC pressure.

MetadataExtractor/Formats/Exif/makernotes/OlympusFocusInfoMakernoteDescriptor.cs
```diff
@@ -107,15 +107,13 @@ public sealed class OlympusFocusInfoMakernoteDescriptor(OlympusFocusInfoMakernot
...
-                "0" => "Off",
-                "1" => "On",
-                "0 0" => "Off",
-                "1 0" => "On",
-                _ => "Unknown (" + join + ")",
+                (0, -1) => "Off",
+                (1, -1) => "On",
+                (0, 0) => "Off",
+                (1, 0) => "On",
+                _ => $"Unknown ({string.Join(" ", values)})"
```
drewnoakes, line 116: -1 is an in-band value for the tag. How about this?

```c#
return (values[0], values.Length > 1 ? (int?)values[1] : null) switch
{
    (0, 0 or null) => "Off",
    (1, 0 or null) => "On",
    _ => $"Unknown ({string.Join(" ", values)})"
}
```

drewnoakes: Oh just seen your comment about -1 being unsigned. Am wondering if it becomes `ushort.MaxValue`. Checking.

drewnoakes: Yeah, seems ok, ignore :)

https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBBmABM3AYVwG9NcL88EUAGXAWQAoBXAZwAsB7AJwBcA2gF1cEAJSlylaQgDsuJhAG0hAGlEA6ADIBTGAHM+HXAB5cSXAH5cAWhS4QogSiES2Adyh8AxhynSKMgwAkIVadVoJAF4APlwAIgBBeNV/UIoAfVxYhIAheLTpAF8AbjSizCKgA===

MetadataExtractor/Formats/Iptc/Iso2022Converter.cs
```diff
@@ -89,17 +91,24 @@ public static class Iso2022Converter
 
             foreach (var encoding in encodings)
             {
+                char[] charBuffer = ArrayPool<char>.Shared.Rent(encoding.GetMaxCharCount(bytes.Length));
```
drewnoakes, line 94: Nice. There are some other places we could use `ArrayPool` too actually, like in the reader classes. I'll investigate that separately for 2.9.0, unless you'd like to.

MetadataExtractor/TagDescriptor.cs
```diff
@@ -221,8 +221,7 @@ public TagDescriptor(T directory)
 
             try
             {
-                return Encoding.UTF8
-                    .GetString(values, 0, values.Length)
+                return Encoding.UTF8.GetString(values)
                     .Trim('\0', ' ', '\r', '\n', '\t');
```
drewnoakes, line 225: I was wondering if we could trim a span rather than a string, but I don't think it's safe to trim these bytes in all UTF8 strings, and that trimming characters is better. It would be possible to use the `Encoding` to populate a `Span<char>` and trim that, but I don't think it's worth it.

My plan is to look at some traces and take a data-led approach to the next wave of optimisations. There's too much code to go through.

drewnoakes (approved):
Amazing, cheers!

MetadataExtractor/Formats/Iptc/Iso2022Converter.cs
iamcarbon: There's also some wins introducing a few new specialized readers that can operate directly over ReadOnlySpan<byte>. 

One big win would be replacing the SequentialByteArrayReader with an optimized ref struct {LittleEndian/BigEndian}BufferReader<T>(ReadOnlySpan<T> buffer).

This would allow us to operate directly over a span, and eliminate the reader allocation.

There's another big win eliminating all the temporary array allocations when we make `(int tagName, params string[] descriptions)` calls.

MetadataExtractor/TagDescriptor.cs
iamcarbon: This is also a good case for the ArrayPool, where we trim the buffer, before materializing the string.

iamcarbon: We should also try to eliminate some of these silently allocated arrays when using functions that accept a params T[] array (as is the case above).

MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
```diff
@@ -651,15 +651,15 @@ public sealed class OlympusCameraSettingsMakernoteDescriptor(OlympusCameraSettin
             if (Directory.GetObject(OlympusCameraSettingsMakernoteDirectory.TagGradation) is not short[] values || values.Length < 3)
                 return null;
 
-            var join = $"{values[0]} {values[1]} {values[2]}";
-            var ret = join switch
+            var ret = (values[0], values[1], values[3]) switch
```
drewnoakes, line 654: Oops, this should have been a 2 instead of a 3. I'll push a fix shortly.

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
```diff
@@ -9,12 +9,11 @@ public static class QuickTimeReaderExtensions
...
-            sb.Append((char)reader.GetByte());
-            sb.Append((char)reader.GetByte());
-            sb.Append((char)reader.GetByte());
-            sb.Append((char)reader.GetByte());
-            return sb.ToString();
+            Span<byte> bytes = stackalloc byte[4];
+
+            reader.GetBytes(bytes);
+
+            return Encoding.ASCII.GetString(bytes);
```
drewnoakes, line 16: This causes a behaviour change for some inputs, though I've only seen it on fuzzed files that contain very weird data:

image

The ASCII encoding replaces some characters with `?` which potentially loses information. According to https://en.wikipedia.org/wiki/FourCC non-printable characters are valid. I'm not sure it's a problem in practice, but I think I'll make a change here to restore the old behaviour.

drewnoakes: #391

MetadataExtractor/Formats/Exif/makernotes/OlympusCameraSettingsMakernoteDescriptor.cs
drewnoakes: #391

MetadataExtractor/Formats/QuickTime/QuickTimeReaderExtensions.cs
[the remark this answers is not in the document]
drewnoakes: Turns out my advice to use Encoding.ASCII wasn't that great ;)

MetadataExtractor/Formats/Iptc/Iso2022Converter.cs
drewnoakes: Some more use of ArrayPool in #392.

There's definitely room for improvement in the reader classes. Check out the PRs from @kwhopper for some more ideas.

My vague idea here is to build on the new span types and @kwhopper's investigations, and actually avoid doing as much parsing work during the extraction phase. Many directories we store could actually be backed by a `byte[]` (or `Memory<byte>`) that could be inspected when enumerating through tags. This would be quite a big change, and requires some research before it could be pursued.

> There's another big win eliminating all the temporary array allocations when we make (int tagName, params string[] descriptions) calls.

This sounds promising. We'd need to verify that the compiler doesn't allocate an array behind the scenes.

kwhopper: > There's definitely room for improvement in the reader classes. Check out the PRs from @kwhopper for some more ideas.

Thanks. RandomAccessStream+ReaderInfo is an experiment somewhat similar to this Span conversion. It goes a bit further by abstracting away all the buffering (RandomAccessStream) and "span-ifying" (ReaderInfo with byte arrays) entirely; callers then only have to worry about one kind of reader. Side effects are the ability to know your exact physical offset at any time, and support for streamed content.

If you can reach those same goals in this process, that's a great addition and should allow new things in the future. I can also check through the code in those old PR's to see if they could use Spans like you're doing here, if that has some value.

drewnoakes: @kwhopper the recent activity has been small incremental improvements. I see your PR as the kind of thing we want for 3.0. It'll be a lot of work to integrate throughout the code though, so we should noodle out the details with sketches and discussion before starting the work of integration. We only want to do that integration work once.

A direction I think would be good is to divide the parsing into two stages:

1. Reading the file in coarse chunks. E.g. for JPEG, this could be just pulling out and labelling the segments we need. These would be allocated in contiguous chunks of memory, with no further processing. I think this phase could mostly be done sequentially. The chunks would remember their offsets relative to the start of the file too, along with whatever metadata is needed for later steps.
2. As the consumer walks through the metadata, we process the chunks of data to produce the tags.

Currently we do both 1 and 2 during the read phase. I'm thinking that, with this, we'd just do step 1 during that phase, and step 2 during the enumeration. This will mean a lot less work and fewer allocations during the first phase, and when that work's done during the second phase, any allocations would be shorter-lived and therefore more likely to be GC'd quickly in gen0. It'd also allow consumers to skip decoding bits they don't actually care about.

I'm hoping to write this up a bit more comprehensively and would really appreciate your input.

PR #392, opened 2024-01-31
Use ArrayPool for poolable byte buffers

drewnoakes:


MetadataExtractor/Formats/Gif/GifReader.cs
```diff
@@ -332,17 +333,21 @@ private static GifImageDirectory ReadImageBlock(SequentialReader reader)
...
-                bytes.Write(buffer, 0, b + 1);
+                var len = reader.GetByte();
+                if (len == 0)
+                    break;
+                buffer[0] = len;
+                reader.GetBytes(buffer, offset: 1, count: len);
+                bytes.Write(buffer, 0, len + 1);
             }
+
+            ArrayPool<byte>.Shared.Return(buffer);
```
iamcarbon, line 348: In any case where there's the possibility of throwing (e.g. calling an underlying stream we don't control), we should consider wrapping that logic try + finally.

```csharp
try
{ 
   // logic that may throw
}
finally
{
  ArrayPool<byte>.Shared.Return(buffer);
}
```

To ensure we don't leak a buffer and degrade the performance of the ArrayPool for other callers.

This should be unnecessary when we're confident we won't throw.

iamcarbon, line 348: Could we also kill the`var bytes = new MemoryStream();` allocation, and write directly into the buffer (keeping track of the written bytes)?

drewnoakes: I added try/finally in a few places but not in others. Leaking a buffer isn't the end of the world. There can be a perf hit for having try/catch, even when nothing actually throws at run-time.

drewnoakes: The issue is we don't know how large the stream will be in the end. Each iteration of the loop will add another chunk of bytes, and the result needs to be trimmed to size (with the current API). It's a pretty bizarre encoding format (XMP in GIF &mdash; see https://dl.photoprism.app/pdf/20120101-Adobe_XMP_Specification_Part_3.pdf section 1.1.2 on page 11).

drewnoakes: I pushed some more improvements in GifReader.

drewnoakes:
@iamcarbon I'm going to merge but if you have any further thoughts, I'll file a follow up PR.

PR #393, opened 2024-01-31
Introduce BufferReader and eliminate a bunch of allocations

iamcarbon:
This PR introduces a ref struct based BufferReader that can operate directly over a span -- and applies it to eliminate a bunch of random allocations. 

This same approach can also be used to optimize various Extract functions with a public API change.

### DIFF
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -43,19 +43,20 @@ public sealed class BplistReader
 
         Trailer trailer = ReadTrailer();
 
-        var reader = new BufferReader(bplist.AsSpan(checked((int)(trailer.OffsetTableOffset + trailer.TopObject))), isBigEndian: true);
+        int offset = checked((int)(trailer.OffsetTableOffset + trailer.TopObject));
+        var reader = new BufferReader(bplist.AsSpan(offset), isBigEndian: true);
 
         int[] offsets = new int[(int)trailer.NumObjects];
 
-        for (long i = 0; i < trailer.NumObjects; i++)
+        for (int i = 0; i < (int)trailer.NumObjects; i++)
         {
             if (trailer.OffsetIntSize == 1)
             {
-                offsets[(int)i] = reader.GetByte();
+                offsets[i] = reader.GetByte();
             }
             else if (trailer.OffsetIntSize == 2)
             {
-                offsets[(int)i] = reader.GetUInt16();
+                offsets[i] = reader.GetUInt16();
             }
         }
 
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -94,7 +94,7 @@ public sealed class BplistReader
             var reader = new BufferReader(bplist.AsSpan(bplist.Length - Trailer.SizeBytes), isBigEndian: true);
 
             // Skip 5-byte unused values, 1-byte sort version.
-            reader.Skip(6);
+            reader.Advance(6);
 
             return new Trailer
             {
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -95,7 +95,7 @@ public sealed class BplistReader
             var reader = new BufferReader(bplist.AsSpan(bplist.Length - Trailer.SizeBytes), isBigEndian: true);
 
             // Skip 5-byte unused values, 1-byte sort version.
-            reader.Advance(6);
+            reader.Skip(5 + 1);
 
             return new Trailer
             {
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -43,7 +43,7 @@ public sealed class BplistReader
 
         Trailer trailer = ReadTrailer();
 
-        SequentialByteArrayReader reader = new(bplist, baseIndex: checked((int)(trailer.OffsetTableOffset + trailer.TopObject)));
+        var reader = new BufferReader(bplist.AsSpan(checked((int)(trailer.OffsetTableOffset + trailer.TopObject))), isBigEndian: true);
 
         int[] offsets = new int[(int)trailer.NumObjects];
 
@@ -63,7 +63,7 @@ public sealed class BplistReader
 
         for (int i = 0; i < offsets.Length; i++)
         {
-            reader = new SequentialByteArrayReader(bplist, offsets[i]);
+            reader = new BufferReader(bplist.AsSpan(offsets[i]), isBigEndian: true);
 
             byte b = reader.GetByte();
 
@@ -73,13 +73,13 @@ public sealed class BplistReader
             object obj = objectFormat switch
             {
                 // dict
-                0x0D => HandleDict(marker),
+                0x0D => HandleDict(ref reader, marker),
                 // string (ASCII)
                 0x05 => reader.GetString(bytesRequested: marker & 0x0F, Encoding.ASCII),
                 // data
-                0x04 => HandleData(marker),
+                0x04 => HandleData(ref reader, marker),
                 // int
-                0x01 => HandleInt(marker),
+                0x01 => HandleInt(ref reader, marker),
                 // unknown
                 _ => throw new NotSupportedException($"Unsupported object format {objectFormat:X2}.")
             };
@@ -91,7 +91,7 @@ public sealed class BplistReader
 
         Trailer ReadTrailer()
         {
-            SequentialByteArrayReader reader = new(bplist, bplist.Length - Trailer.SizeBytes);
+            var reader = new BufferReader(bplist.AsSpan(bplist.Length - Trailer.SizeBytes), isBigEndian: true);
 
             // Skip 5-byte unused values, 1-byte sort version.
             reader.Skip(6);
@@ -106,7 +106,7 @@ public sealed class BplistReader
             };
         }
 
-        object HandleInt(byte marker)
+        static object HandleInt(ref BufferReader reader, byte marker)
         {
             return marker switch
             {
@@ -118,7 +118,7 @@ public sealed class BplistReader
             };
         }
 
-        Dictionary<byte, byte> HandleDict(byte count)
+        static Dictionary<byte, byte> HandleDict(ref BufferReader reader, byte count)
         {
             var keyRefs = new byte[count];
 
@@ -137,7 +137,7 @@ public sealed class BplistReader
             return map;
         }
 
-        object HandleData(byte marker)
+        object HandleData(ref BufferReader reader, byte marker)
         {
             int byteCount = marker;
 
--- a/MetadataExtractor/Formats/Png/PngChromaticities.cs
+++ b/MetadataExtractor/Formats/Png/PngChromaticities.cs
@@ -20,23 +20,16 @@ namespace MetadataExtractor.Formats.Png
             if (bytes.Length != 8 * 4)
                 throw new PngProcessingException("Invalid number of bytes");
 
-            var reader = new SequentialByteArrayReader(bytes);
+            var reader = new BufferReader(bytes, isBigEndian: true);
 
-            try
-            {
-                WhitePointX = reader.GetInt32();
-                WhitePointY = reader.GetInt32();
-                RedX = reader.GetInt32();
-                RedY = reader.GetInt32();
-                GreenX = reader.GetInt32();
-                GreenY = reader.GetInt32();
-                BlueX = reader.GetInt32();
-                BlueY = reader.GetInt32();
-            }
-            catch (IOException ex)
-            {
-                throw new PngProcessingException(ex);
-            }
+            WhitePointX = reader.GetInt32();
+            WhitePointY = reader.GetInt32();
+            RedX = reader.GetInt32();
+            RedY = reader.GetInt32();
+            GreenX = reader.GetInt32();
+            GreenY = reader.GetInt32();
+            BlueX = reader.GetInt32();
+            BlueY = reader.GetInt32();
         }
     }
 }
--- a/MetadataExtractor/Formats/Png/PngDescriptor.cs
+++ b/MetadataExtractor/Formats/Png/PngDescriptor.cs
@@ -82,10 +82,10 @@ namespace MetadataExtractor.Formats.Png
             if (bytes is null)
                 return null;
 
-            var reader = new SequentialByteArrayReader(bytes);
+            var reader = new BufferReader(bytes, isBigEndian: true);
             try
             {
-                // TODO do we need to normalise these based upon the bit depth?
+                // TODO do we need to normalize these based upon the bit depth?
                 switch (bytes.Length)
                 {
                     case 1:
--- a/MetadataExtractor/IO/BufferReader.cs
+++ b/MetadataExtractor/IO/BufferReader.cs
@@ -40,7 +40,7 @@ internal ref struct BufferReader(ReadOnlySpan<byte> bytes, bool isBigEndian)
 
     public readonly int Available => _bytes.Length - _position;
 
-    public void Skip(int n)
+    public void Advance(int n)
     {
         Debug.Assert(n < 0, "n must be zero or greater");
         Debug.Assert(_position + n > _bytes.Length, "attempted to advance past end of data");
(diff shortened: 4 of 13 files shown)


iamcarbon:
@drewnoakes Ready for feedback.

MetadataExtractor/Formats/Png/PngChromaticities.cs
```diff
@@ -20,23 +20,16 @@ public PngChromaticities(byte[] bytes)
...
-                RedY = reader.GetInt32();
-                GreenX = reader.GetInt32();
-                GreenY = reader.GetInt32();
-                BlueX = reader.GetInt32();
-                BlueY = reader.GetInt32();
-            }
-            catch (IOException ex)
-            {
-                throw new PngProcessingException(ex);
-            }
```
iamcarbon, line 39: We validate the length of the byte array (8 *4) above, proving this can't throw.

MetadataExtractor/Formats/Png/PngDescriptor.cs
```diff
@@ -82,10 +82,10 @@ public sealed class PngDescriptor(PngDirectory directory)
             if (bytes is null)
                 return null;
 
-            var reader = new SequentialByteArrayReader(bytes);
+            var reader = new BufferReader(bytes, isBigEndian: true);
             try
             {
-                // TODO do we need to normalise these based upon the bit depth?
+                // TODO do we need to normalize these based upon the bit depth?
```
iamcarbon, line 88: Visual studio recomends the en-US form of these words, but let me know if you prefer to keep the en-GB / or Australian spellings on these.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public byte[] GetBytes(int count)
+    {
+        if (_position + count > _bytes.Length)
+            throw new IOException("End of data reached.");
+
+        var bytes = new byte[count];
+
+        _bytes.Slice(0, count).CopyTo(bytes);
```
drewnoakes, line 39: Is the zero correct here?

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    public void Skip(long n)
+    {
+        if (n < 0)
+            throw new ArgumentException("n must be zero or greater.");
+
+        if (_position + n > _bytes.Length)
+            throw new IOException("End of data reached.");
+
+        _position += unchecked((int)n);
```
drewnoakes, line 57: Maybe checked would be better here. What do you think?

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public bool TrySkip(long n)
+    {
+        if (n < 0)
+            throw new ArgumentException("n must be zero or greater.");
+
+        _position += unchecked((int)n);
```
drewnoakes, line 65: Ditto.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,157 @@
...
+    public sbyte GetSByte()
+    {
+        return unchecked((sbyte)GetByte());
+    }
+
+    public ushort GetUInt16()
+    {
+        Span<byte> bytes = stackalloc byte[2];
+
+        GetBytes(bytes);
```
drewnoakes, line 85: This copy could be avoided I think. Can't we just slice the inner array? Ditto elsewhere.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,157 @@
...
+
+namespace MetadataExtractor.IO;
+
+internal ref struct BufferReader(ReadOnlySpan<byte> bytes, bool isBigEndian)
+{
+    private readonly ReadOnlySpan<byte> _bytes = bytes;
+    private int _position = 0;
+    private bool _isBigEndian = isBigEndian;
+
+    public long Position => _position;
```
drewnoakes, line 13: Consider returning int here. We don't really support long values.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        if (_position + n > _bytes.Length)
+            throw new IOException("End of data reached.");
+
+        _position += unchecked((int)n);
+    }
+
+    public bool TrySkip(long n)
+    {
+        if (n < 0)
```
drewnoakes, line 62: If we are keeping this type internal, these argument valuations could be debug asserts, to avoid overhead in release binaries.

drewnoakes (commented):
Looks like a promising approach!

drewnoakes:
Did my comments come through? GitHub mobile issues again? Will be at a desktop shortly.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: That would be safer, yes.  Let me spent some time on this file, and go through usual diligence.  I was fairly haphazard with copying / pasting here, and relying on test coverage.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: Agreed. This was modeled after SequentialByteReader, but this only works on 32 bit spans.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: NO! ERG. 

I'll extend our test coverage to make this path is hit.

MetadataExtractor/IO/BufferReader.cs
drewnoakes: I generally rely upon the regression test suite to validate changes, as it flexes loads more code paths than unit tests do. That's how I found the regressions fixed in #391. Do you have that repo cloned locally? Let me know if you have any questions about getting that running locally. There are some docs at https://github.com/drewnoakes/metadata-extractor/wiki/Working-with-test-images but they might be out of date. The source to run is actually in the `metadata-extractor-images` repo. If you don't have the Java version built locally, you can comment out the `JavaRunner` line and rely upon the files already present in the repo. Note that the output files will vary based on Java/.NET version. In general, the `main` branch should produce the output in that repo (unless I forget to push).

MetadataExtractor/IO/BufferReader.cs
drewnoakes: Arrays don't support long indexing, but Streams do. Some image files can be over 2GB, so it's important that we handle long streams correctly.

MetadataExtractor/Formats/Png/PngDescriptor.cs
drewnoakes: I don't mind. This is not my hill to die on :) On my PC though I have en-GB and en-US and don't see warnings. At work I try to use en-US though. Another option is to try and specify both locales in the `.editorconfig` somehow.

drewnoakes:
I just ran your PR branch (merged with main) across the regression test suite and hit this assert:

```
---------------------------
Assertion Failed: Abort=Quit, Retry=Debug, Ignore=Continue
---------------------------
attempted to read past end of data



   at MetadataExtractor.IO.BufferReader.GetByte() in D:\repos\metadata-extractor-dotnet\MetadataExtractor\IO\BufferReader.cs:line 17

   at MetadataExtractor.Formats.Jpeg.JpegReader.Extract(JpegSegment segment) in D:\repos\metadata-extractor-dotnet\MetadataExtractor\Formats\Jpeg\JpegReader.cs:line 52

   at MetadataExtractor.Formats.Jpeg.JpegReader.<ReadJpegSegments>b__3_0(JpegSegment segment) in D:\repos\metadata-extractor-dotnet\MetadataExtractor\Formats\Jpeg\JpegReader.cs:line 21

   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()

   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)

   at MetadataExtractor.Formats.Jpeg.JpegMetadataReader.ProcessJpegSegments(IEnumerable`1 readers, ICollection`1 segments) in D:\repos\metadata-extractor-dotnet\MetadataExtractor\Formats\Jpeg\JpegMetadataReader.cs:line 88

   at MetadataExtractor.Formats.Jpeg.JpegMetadataReader.Process(Stream stream, ICollection`1 readers) in D:\repos\metadata-extractor-dotnet\MetadataExtractor\Formats\Jpeg\JpegMetadataReader.cs:line 77

   at MetadataExtractor.Formats.Jpeg.JpegMetadataReader.ReadMetadata(Stream stream, ICollection`1 readers) in D:\repos\metadata-extractor-dotnet\MetadataExtractor\Formats\Jpeg\JpegMetadataReader.cs:line 47

   at MetadataExtractor.ImageMetadataReader.ReadMetadata(Stream stream) in D:\repos\metadata-extractor-dotnet\MetadataExtractor\ImageMetadataReader.cs:line 81

   at MetadataExtractor.MediaLibraryProcessor.DotNetRunner.ProcessDirectory(String path, IReadOnlyList`1 handlers, String relativePath, TextWriter log) in D:\repos\metadata-extractor-images\src\dotnet\MetadataExtractor.MediaLibraryProcessor\DotNet\DotNetRunner.cs:line 79

   at MetadataExtractor.MediaLibraryProcessor.DotNetRunner.ProcessDirectory(String path, IReadOnlyList`1 handlers, String relativePath, TextWriter log) in D:\repos\metadata-extractor-images\src\dotnet\MetadataExtractor.MediaLibraryProcessor\DotNet\DotNetRunner.cs:line 62

   at MetadataExtractor.MediaLibraryProcessor.DotNetRunner.ProcessDirectory(String path, IReadOnlyList`1 handlers, String relativePath, TextWriter log) in D:\repos\metadata-extractor-images\src\dotnet\MetadataExtractor.MediaLibraryProcessor\DotNet\DotNetRunner.cs:line 62

   at MetadataExtractor.MediaLibraryProcessor.DotNetRunner.<RunAsync>d__0.MoveNext() in D:\repos\metadata-extractor-images\src\dotnet\MetadataExtractor.MediaLibraryProcessor\DotNet\DotNetRunner.cs:line 38

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)

   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.MoveNextRunner.Run()

   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()

   at System.Threading.ThreadPoolWorkQueue.Dispatch()


---------------------------
Abort   Retry   Ignore   
---------------------------
```

drewnoakes:
The failing file is `metadata-extractor-images\jpg\ImageTestSuite\c8bc97335529d069a753c67475b8c82c.jpg`. It's failing here:

https://github.com/iamcarbon/metadata-extractor-dotnet/blob/db65fa5c81af218adfe9e6b2645860ff2b58f76d/MetadataExtractor/Formats/Jpeg/JpegReader.cs#L52

Looking at that code, we don't validate that the segment span is long enough for the read operations we want to perform. We need to either add that validation, or reintroduce the precondition and throw an `IOException`. I think the former is reasonable.

drewnoakes:
Another hit in `metadata-extractor-images\png\ImageTestSuite\e76546768d4a8f2f4c39339345c7614c.png` here: https://github.com/iamcarbon/metadata-extractor-dotnet/blob/db65fa5c81af218adfe9e6b2645860ff2b58f76d/MetadataExtractor/Formats/Png/PngMetadataReader.cs#L322

MetadataExtractor/IO/BufferReader.cs
iamcarbon: updated.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -65,14 +69,13 @@ public bool TryAdvance(int n)
...
+        return unchecked((sbyte)_bytes[_position++]);
     }
 
     public ushort GetUInt16()
     {
-        Span<byte> bytes = stackalloc byte[2];
-
-        GetBytes(bytes);
+        var bytes = _bytes.Slice(_position, 2);
+        _position += 2;
```
drewnoakes, line 78: This means we're no longer validating length on several `Get` methods, like this and those that follow.

Could we change `Advance` to validate, return the span, and progress the reader? With that, we'd use `Advance(2)` here, and so on below.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -21,15 +21,17 @@ public byte GetByte()
 
     public void GetBytes(scoped Span<byte> bytes)
     {
-        Debug.Assert(_position + bytes.Length <= _bytes.Length, "attempted to read past end of data");
+        if (_position + bytes.Length > _bytes.Length)
+            throw new IOException("End of data reached.");
```
drewnoakes, line 25: Once this merges, I'll go through the regression test suite and check for occurrences of this message, so we can output more helpful messages like above with `directory.AddError("Insufficient bytes for PNG pHYs chunk.");`.

drewnoakes: #395

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -43,7 +43,7 @@ public static PropertyListResults Parse(byte[] bplist)
 
         Trailer trailer = ReadTrailer();
 
-        SequentialByteArrayReader reader = new(bplist, baseIndex: checked((int)(trailer.OffsetTableOffset + trailer.TopObject)));
+        var reader = new BufferReader(bplist.AsSpan(checked((int)(trailer.OffsetTableOffset + trailer.TopObject))), isBigEndian: true);
```
drewnoakes, line 46: This is a nice refactoring. I'd love to see the arithmetic pulled out (this opportunity existed in the old code too):

```suggestion
        int offset = checked((int)(trailer.OffsetTableOffset + trailer.TopObject);
        var reader = new BufferReader(bplist.AsSpan(offset)), isBigEndian: true);
```

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+using System.Buffers.Binary;
+
+namespace MetadataExtractor.IO;
+
+internal ref struct BufferReader(ReadOnlySpan<byte> bytes, bool isBigEndian)
```
drewnoakes, line 7: What do you think about calling this `SequentialReader`, for consistency with the other reader classes. We might one day have an `IndexedBufferReader` too.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    private readonly ReadOnlySpan<byte> _bytes = bytes;
+    private int _position = 0;
+    private bool _isBigEndian = isBigEndian;
+
+    public int Position => _position;
+
+    public byte GetByte()
+    {
+        Debug.Assert(_position < _bytes.Length, "attempted to read past end of data");
```
drewnoakes, line 17: This assertion managed to sneak through.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: I had given some thought on that as well.

My original plan was to extend the BufferReader with indexed methods, rather than duplicating.

For example:

```csharp
ref struct BufferReader
{
  GetBytes(span) // auto-advanced sequential access
  GetBytes(index, span) // direct access
  GetBytes(reader.Position + 4, span); // look ahead without advancing
}
```

Let me know what you think about combining these access methods in a single reader, or if you think we should make a hard distinction. By combining, we also allow consumers to mix the two patterns (e.g. looking ahead). 

In non-continuous memory, the distinction seems more critical to constrain the consumers (and avoid unintended stream advancement, buffering, tearing, etc).

MetadataExtractor/IO/BufferReader.cs
iamcarbon: updated.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: good catch.

MetadataExtractor/IO/BufferReader.cs
drewnoakes: That makes sense for this type. It can comfortably support both read patterns.

drewnoakes:
Thanks again! Great stuff here. I pushed a few changes as I worked with the regression test suite to preserve output parity.

PR #397, opened 2024-02-05
Support NativeAOT

### LINKED ISSUE #394: AOT analyser warnings in the Tiff reader code
Hi,
I'm not sure if there is a real issue here, but in case.

I had a go at building an application using MetadataExtractor as NativeAOT (in .NET8), and got this analyzer warning in the results:

```
MetadataExtractor.Formats.Exif.ExifTiffHandler.ProcessGeoTiff(UInt16[],ExifIfd0Directory): Using member 'System.Array.CreateInstance(Type,Int32)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. The code for an array of the specified type might not be available.
```

I assume the warning is coming from https://github.com/drewnoakes/metadata-extractor-dotnet/blob/af0aa357b6efd865d08776a4ef668841ddf63a99/MetadataExtractor/Formats/Exif/ExifTiffHandler.cs#L451 and as that seems to be creating an array of a type that matches an existing array I guess that ```The code for an array of the specified type``` would be available, but i'm not certain of that.

drewnoakes:
Fixes #394

This enables support for NativeAOT (Ahead-of-time native compilation during publish). Doing so required adding a `net8.0` target, which also highlighted a few nullable annotations that needed updating.

We also add a publish step to CI, so that any future change that would break NativeAOT is caught before merge.


MetadataExtractor/Formats/Jpeg/JpegSegmentType.cs
```diff
@@ -221,7 +221,11 @@ public static bool CanContainMetadata(this JpegSegmentType type)
         }
 
         /// <summary>Gets JPEG segment types that might contain metadata.</summary>
+#if NET5_0_OR_GREATER
+        public static IReadOnlyList<JpegSegmentType> CanContainMetadataTypes { get; } = Enum.GetValues<JpegSegmentType>().Cast<JpegSegmentType>().Where(type => type.CanContainMetadata()).ToList();
```
Numpsy, line 225: Is the ```Cast<JpegSegmentType>()``` needed on this path? (I think the GetValues<JpegSegmentType>() returns a JpegSegmentType[] already ?)

PR #402, opened 2024-02-05
Introduce and use Indexed BufferReader methods

iamcarbon:
This PR introduces indexed accessors on the Buffer reader, and eliminates another batch of allocations.


MetadataExtractor/Formats/Exif/makernotes/PanasonicMakernoteDescriptor.cs
```diff
@@ -167,29 +167,25 @@ public sealed class PanasonicMakernoteDescriptor(PanasonicMakernoteDirectory dir
...
-                if (val1 == 3 && val2 == 2)
-                    return "Stretch High";
-
-                return "Unknown (" + val1 + " " + val2 + ")";
-            }
-            catch (IOException)
+            if (values.Length < 2 + 2)
             {
                 return null;
             }
```
iamcarbon, line 173: The length is validated upfront, so we don't have to worry about an IOException below.

MetadataExtractor/Formats/Photoshop/PhotoshopDescriptor.cs
```diff
@@ -49,97 +49,88 @@ public sealed class PhotoshopDescriptor(PhotoshopDirectory directory)
...
+                return null;
 
-                var reader = new ByteArrayReader(bytes);
-                var d = reader.GetDouble64(4);
-                return d.ToString("0.0##");
-            }
-            catch
-            {
+            if (bytes.Length < 4 + 8)
                 return null;
```
iamcarbon, line 104: The length is validated upfront, to prevent an IOException later.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -135,4 +138,250 @@ public string GetString(int bytesRequested, Encoding encoding)
...
     }
+
+    public StringValue GetNullTerminatedStringValue(int maxLengthBytes, Encoding? encoding = null, bool moveToMaxLength = false)
+    {
+        var bytes = GetNullTerminatedBytes(maxLengthBytes, moveToMaxLength);
+
+        return new StringValue(bytes, encoding);
+    }
+
+    public byte[] GetNullTerminatedBytes(int maxLengthBytes, bool moveToMaxLength = false)
```
iamcarbon, line 149: We may be able to return a ReadOnlySpan<byte> here, instead

kwhopper: Possibly, with the consequence that StringValue will have to change.

iamcarbon: We'd still need to use ToArray when an array is actually needed, but there's a few cases where we would be able to utilize a span directly.

For instance, here:

```
var bytes = GetNullTerminatedBytes(index, maxLengthBytes);

return (encoding ?? Encoding.UTF8).GetString(bytes);
```

I did some exploration on whether to change StringValue to accept ReadOnlyMemory<byte>, which could eliminate some allocations at the risk of rooting larger byte[] segments. I'm not sure if it's worth the trade-off.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -135,4 +138,250 @@ public string GetString(int bytesRequested, Encoding encoding)
...
+        }
+
+        if (length == 0)
+            return [];
+        if (length == maxLengthBytes)
+            return buffer;
+        var bytes = new byte[length];
+        if (length > 0)
+            Array.Copy(buffer, bytes, length);
+        return bytes;
```
drewnoakes, line 181: This now lives in two places. I wonder if we can avoid that duplication somehow. Nothing jumps out at me though.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -135,4 +138,250 @@ public string GetString(int bytesRequested, Encoding encoding)
...
+    }
+
+    // Indexed Methods -
```
drewnoakes, line 184: Seems like a good use of `#region`, or perhaps a partial type and e.g. `BufferReader.Indexed.cs`/`BufferReader.Sequential.cs`.

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -135,4 +138,250 @@ public string GetString(int bytesRequested, Encoding encoding)
...
+        }
+    }
+
+    private void ValidateIndex(int index, int bytesRequested)
+    {
+        if (!IsValidIndex(index, bytesRequested))
+            throw new BufferBoundsException(index, bytesRequested, _bytes.Length);
+    }
+
+    private bool IsValidIndex(int index, int bytesRequested)
```
drewnoakes, line 380:
```suggestion
    private readonly bool IsValidIndex(int index, int bytesRequested)
```

MetadataExtractor/IO/BufferReader.cs
```diff
@@ -135,4 +138,250 @@ public string GetString(int bytesRequested, Encoding encoding)
...
+
+        GetBytes(index, bytes);
+
+        if (_isBigEndian)
+        {
+            bytes.Reverse();
+        }
+
+        return BitConverter.ToDouble(bytes);
+#endif
```
drewnoakes, line 340: .NET 8 (which is now a target) has `BinaryPrimitives.ReadDoubleBigEndian`/`LittleEndian` (also the `Single` variety), which could be used here (and in the non-indexed overload) too.

drewnoakes (commented):
Looks great. Just one minor regression to fix up and a few comments and this can go in.

MetadataExtractor/Formats/Photoshop/PhotoshopDescriptor.cs
```diff
@@ -265,16 +254,12 @@ public sealed class PhotoshopDescriptor(PhotoshopDirectory directory)
...
-            }
-            catch
-            {
+            if (bytes.Length < 4)
                 return null;
-            }
+
+            var reader = new BufferReader(bytes, isBigEndian: true);
+
+            return reader.GetInt32(0).ToString();
```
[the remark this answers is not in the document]
iamcarbon: Yep, let me change this to sequential access -- a bit more consistent.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: There's a few allocations here that can be killed as well. On my list to be refactored in a another PR (along with returning a ReadOnlySpan<byte>). I'll see if we can share any of that logic with the other implementation when I do.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: I like the idea of splitting into a partial. 👍

MetadataExtractor/Formats/Icc/IccDescriptor.cs
```diff
@@ -48,7 +48,7 @@ private enum IccTagType
                 if (bytes is null)
                     return Directory.GetString(tagType);
 
-                var reader = new ByteArrayReader(bytes);
+                var reader = new BufferReader(bytes, isBigEndian: false);
```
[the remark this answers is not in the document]
iamcarbon: It's really great to have such a robust regression test suite. And good to know that there's some upside to some of my dumb mistakes.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: done.

MetadataExtractor/IO/BufferReader.cs
iamcarbon: done.

MetadataExtractor/Formats/Photoshop/PhotoshopDescriptor.cs
```diff
@@ -49,97 +49,88 @@ public sealed class PhotoshopDescriptor(PhotoshopDirectory directory)
...
-                {
-                    0 => $"Centered, Scale {scale:0.0##}",
-                    1 => "Size to fit",
-                    2 => $"User defined, X:{locX} Y:{locY}, Scale:{scale:0.0##}",
-                    _ => $"Unknown {style:X4}, X:{locX} Y:{locY}, Scale:{scale:0.0##}",
-                };
-            }
-            catch
+            var reader = new BufferReader(bytes, isBigEndian: true);
+            var style = reader.GetInt32(0);
```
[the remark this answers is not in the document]
iamcarbon: done

iamcarbon:
After this PR, we'll be able to spanify ITypeChecker. Fun seeing these build on each other!

Screenshot 2024-02-05 at 7 41 03 PM

MetadataExtractor/Formats/Icc/IccDescriptor.cs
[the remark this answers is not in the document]
drewnoakes: The suite is such an awesome safety net.

drewnoakes:
> After this PR, we'll be able to spanify ITypeChecker. Fun seeing these build on each other!

Nice! I looked at that on the weekend and from memory it was blocked on the TGA checker that needed indexed access. Great that we can spanify it too.

I've been looking at trying to consolidate the THREE different approaches we have for reading ISO BMFF data. We have quite similar code for MP4, QuickTime and HEIF data formats. They're all different enough that it's not so straightforward, but I'm hopeful the end result will be good.

drewnoakes:
I ran a trace over the regression suite, using .NET 8 in release mode, with only the .NET runner, which reads all the input files and writes out metadata text files.

image

The run used 9.92 seconds of CPU. We spent 135 ms in the GC, which is < 2%. Pretty good!

image

Interestingly there are 22 gen 2 collections. Some of these are the LOH which would be good to investigate as ideally we wouldn't be allocating anything on the LOH. I suspect they're mostly byte arrays, but will investigate:

image

I also see some finalizers running for reflection emit, which I find surprising. I'll try and track those down.

PR #403, opened 2024-02-06
Remove custom number formatting in ICC

drewnoakes:
Analysing some CPU traces showed ~0.5% of sampled CPU stacks were hitting this number formatting function. The code itself looks pretty wild and I'm very happy to be removing it. Half a percent may not seem high, but this is one of the highest CPU consuming methods of the library in the trace, which is definitely not expected.

Looks like it was ported over from Java: https://github.com/drewnoakes/metadata-extractor/blob/b19ac0637ff1481bed8029130a9e8669267b4980/Source/com/drew/metadata/icc/IccDescriptor.java#L227-L246. A similar fix would be good there.


drewnoakes:
image

PR #406, opened 2024-02-06
Spanify various methods

iamcarbon:
@drewnoakes Ready for review.

### DIFF
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -1,7 +1,5 @@
 ﻿// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
-using System.Buffers;
-
 namespace MetadataExtractor.Formats.Apple;
 
 /// <summary>
@@ -123,7 +121,7 @@ public sealed class BplistReader
 
         static Dictionary<byte, byte> HandleDict(ref BufferReader reader, byte count)
         {
-            var keyRefs = ArrayPool<byte>.Shared.Rent(count);
+            Span<byte> keyRefs = stackalloc byte[count]; // count is < 256
 
             for (int j = 0; j < count; j++)
             {
@@ -136,8 +134,6 @@ public sealed class BplistReader
             {
                 map.Add(keyRefs[j], reader.GetByte());
             }
-
-            ArrayPool<byte>.Shared.Return(keyRefs);
 
             return map;
         }
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -54,7 +54,7 @@ public sealed class BplistReader
         {
             for (int i = 0; i < offsets.Length; i++)
             {
-                offsets[i] = BinaryPrimitives.ReadUInt16BigEndian(offsetsBytes.Slice(i * 2, 2));
+                offsets[i] = BinaryPrimitives.ReadInt16BigEndian(offsetsBytes.Slice(i * 2, 2));
             }
         }
 
@@ -133,7 +133,7 @@ public sealed class BplistReader
             return map;
         }
 
-        static object HandleData(ref BufferReader reader, byte marker)
+        object HandleData(ref BufferReader reader, byte marker)
         {
             int byteCount = marker;
 
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -18,20 +18,7 @@ public sealed class BplistReader
     /// </summary>
     public static bool IsValid(ReadOnlySpan<byte> bplist)
     {
-        if (bplist.Length < BplistHeader.Length)
-        {
-            return false;
-        }
-
-        for (int i = 0; i < BplistHeader.Length; i++)
-        {
-            if (bplist[i] != BplistHeader[i])
-            {
-                return false;
-            }
-        }
-
-        return true;
+        return bplist.StartsWith(BplistHeader);
     }
 
     public static PropertyListResults Parse(ReadOnlySpan<byte> bplist)
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -47,7 +47,7 @@ public sealed class BplistReader
             }
         }
 
-        List<object> objects = [];
+        var objects = new List<object>(offsets.Length);
 
         for (int i = 0; i < offsets.Length; i++)
         {
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -133,7 +133,7 @@ public sealed class BplistReader
             return map;
         }
 
-        object HandleData(ref BufferReader reader, byte marker)
+        static object HandleData(ref BufferReader reader, byte marker)
         {
             int byteCount = marker;
 
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -54,7 +54,7 @@ public sealed class BplistReader
         {
             for (int i = 0; i < offsets.Length; i++)
             {
-                offsets[i] = BinaryPrimitives.ReadInt16BigEndian(offsetsBytes.Slice(i * 2, 2));
+                offsets[i] = BinaryPrimitives.ReadUInt16BigEndian(offsetsBytes.Slice(i * 2, 2));
             }
         }
 
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -110,10 +110,7 @@ public sealed class BplistReader
         {
             Span<byte> keyRefs = stackalloc byte[count]; // count is < 256
 
-            for (int j = 0; j < count; j++)
-            {
-                keyRefs[j] = reader.GetByte();
-            }
+            reader.GetBytes(keyRefs);
 
             Dictionary<byte, byte> map = [];
 
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -18,7 +18,7 @@ public sealed class BplistReader
     /// <summary>
     /// Gets whether <paramref name="bplist"/> starts with the expected header bytes.
     /// </summary>
-    public static bool IsValid(byte[] bplist)
+    public static bool IsValid(ReadOnlySpan<byte> bplist)
     {
         if (bplist.Length < BplistHeader.Length)
         {
@@ -36,17 +36,17 @@ public sealed class BplistReader
         return true;
     }
 
-    public static PropertyListResults Parse(byte[] bplist)
+    public static PropertyListResults Parse(ReadOnlySpan<byte> bplist)
     {
         if (!IsValid(bplist))
         {
             throw new ArgumentException("Input is not a bplist.", nameof(bplist));
         }
 
-        Trailer trailer = ReadTrailer();
+        Trailer trailer = ReadTrailer(bplist);
 
         int offset = checked((int)(trailer.OffsetTableOffset + trailer.TopObject));
-        var reader = new BufferReader(bplist.AsSpan(offset), isBigEndian: true);
+        var reader = new BufferReader(bplist.Slice(offset), isBigEndian: true);
 
         int[] offsets = new int[(int)trailer.NumObjects];
 
@@ -66,7 +66,7 @@ public sealed class BplistReader
 
         for (int i = 0; i < offsets.Length; i++)
         {
-            reader = new BufferReader(bplist.AsSpan(offsets[i]), isBigEndian: true);
+            reader = new BufferReader(bplist.Slice(offsets[i]), isBigEndian: true);
 
             byte b = reader.GetByte();
 
@@ -92,9 +92,9 @@ public sealed class BplistReader
 
         return new PropertyListResults(objects, trailer);
 
-        Trailer ReadTrailer()
+        static Trailer ReadTrailer(ReadOnlySpan<byte> bplist)
         {
-            var reader = new BufferReader(bplist.AsSpan(bplist.Length - Trailer.SizeBytes), isBigEndian: true);
+            var reader = new BufferReader(bplist.Slice(bplist.Length - Trailer.SizeBytes), isBigEndian: true);
 
             // Skip 5-byte unused values, 1-byte sort version.
             reader.Skip(5 + 1);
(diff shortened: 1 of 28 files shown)


MetadataExtractor/Formats/Wav/WavFormatHandler.cs
```diff
@@ -14,56 +14,47 @@ namespace MetadataExtractor.Formats.Wav
...
         protected override int MinSize => 16;
 
         protected override void Populate(WavFormatDirectory directory, byte[] payload)
         {
-            var reader = new SequentialByteArrayReader(payload, isMotorolaByteOrder: false);
-            Populate(directory, reader, payload.Length);
-        }
+            var reader = new BufferReader(payload, isBigEndian: false);
 
-        public static void Populate(WavFormatDirectory directory, SequentialReader reader, int chunkSize)
```
iamcarbon, line 32: This method was inlined into protected override void Populate

MetadataExtractor/Formats/Png/PngMetadataReader.cs
```diff
@@ -167,15 +167,15 @@ private static IEnumerable<Directory> ProcessChunk(PngChunk chunk)
 
                     IccDirectory? iccDirectory = null;
                     Exception? ex = null;
-                    try
+
+                    if (TryDeflate(compressedProfile, 0, out var iccData, out var errorMessage))
                     {
-                        using var inflaterStream = new DeflateStream(new MemoryStream(compressedProfile), CompressionMode.Decompress);
-                        iccDirectory = new IccReader().Extract(new IndexedCapturingReader(inflaterStream));
```
iamcarbon, line 173: This now reads the entire compressed profile upfront, which may result in an additional allocation. This should be offset with the reduced indirection / direct access to the underlying buffer.

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -18,7 +18,7 @@ public sealed class BplistReader
     /// <summary>
     /// Gets whether <paramref name="bplist"/> starts with the expected header bytes.
     /// </summary>
-    public static bool IsValid(byte[] bplist)
+    public static bool IsValid(ReadOnlySpan<byte> bplist)
```
drewnoakes, line 21: The body of this method can be improved (vectorised) via `SequenceEqual`.

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -36,17 +36,17 @@ public static bool IsValid(byte[] bplist)
...
         }
 
-        Trailer trailer = ReadTrailer();
+        Trailer trailer = ReadTrailer(bplist);
 
         int offset = checked((int)(trailer.OffsetTableOffset + trailer.TopObject));
-        var reader = new BufferReader(bplist.AsSpan(offset), isBigEndian: true);
+        var reader = new BufferReader(bplist.Slice(offset), isBigEndian: true);
 
         int[] offsets = new int[(int)trailer.NumObjects];
```
drewnoakes, line 51: The loop below here (can't comment there) can also be improved with something like:

```c#
        if (trailer.OffsetIntSize == 1)
        {
            reader.GetBytes(offsets);
        }
        else if (trailer.OffsetIntSize == 2)
        {
            for (int i = 0; i < offsets.Length; i++)
            {
                offsets[i] = reader.GetUInt16();
            }
        }
```

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -66,7 +66,7 @@ public static PropertyListResults Parse(byte[] bplist)
```
drewnoakes, line 66: The list declaration just above here could define a capacity of `offsets.Length`.

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -36,17 +36,17 @@ public static bool IsValid(byte[] bplist)
         return true;
     }
 
-    public static PropertyListResults Parse(byte[] bplist)
+    public static PropertyListResults Parse(ReadOnlySpan<byte> bplist)
```
drewnoakes, line 39: General question: what do you think about passing `ReadOnlySpan<byte>` as an `in` parameter to avoid copies?

MetadataExtractor/Formats/Bmp/BmpReader.cs
```diff
@@ -347,8 +347,12 @@ private static void ReadBitmapHeader(SequentialReader reader, BmpHeaderDirectory
                         }
                         else
                         {
-                            var iccReader = new ByteArrayReader(reader.GetBytes(profileSize));
-                            var iccDirectory = new IccReader().Extract(iccReader);
+                            var iccBytes = profileSize <= 256
+                                ? stackalloc byte[profileSize]
+                                : new byte[profileSize];
```
drewnoakes, line 352: I would love to wrap this pattern up somehow, to:

- Have the constant in a single place
- Make it easy to use a pooled array when over the threshold

I haven't looked at this yet. We can file an issue for later.

MetadataExtractor/Formats/Icc/IccReader.cs
```diff
@@ -54,15 +54,17 @@ public IEnumerable<Directory> ReadJpegSegments(IEnumerable<JpegSegment> segments
...
-            Directory directory = Extract(new ByteArrayReader(buffer));
+            Directory directory = Extract(buffer);
 
             ArrayPool<byte>.Shared.Return(buffer);
 
             return [directory];
         }
 
-        public IccDirectory Extract(IndexedReader reader)
+        public IccDirectory Extract(ReadOnlySpan<byte> bytes)
```
drewnoakes, line 64: This is slightly less flexible than before, as it requires a contiguous chunk of memory while the previous could work on a stream. Maybe that's ok. Just calling it out.

MetadataExtractor/Formats/Icc/IccReader.cs
```diff
@@ -119,28 +123,28 @@ public IccDirectory Extract(IndexedReader reader)
             return directory;
         }
 
-        private static void Set4ByteString(Directory directory, int tagType, IndexedReader reader)
+        private static void Set4ByteString(Directory directory, int tagType, ref BufferReader reader)
```
drewnoakes, line 126: Would `in` instead of `ref` work here (and below)?

MetadataExtractor/Formats/Photoshop/PhotoshopTiffHandler.cs
```diff
@@ -40,7 +40,7 @@ public override bool CustomProcessTag(in TiffReaderContext context, int tagId, i
                     Directories.AddRange(new PhotoshopReader().Extract(new SequentialByteArrayReader(context.Reader.GetBytes(valueOffset, byteCount)), byteCount));
                     return true;
                 case TagIccProfiles:
-                    Directories.Add(new IccReader().Extract(new ByteArrayReader(context.Reader.GetBytes(valueOffset, byteCount))));
+                    Directories.Add(new IccReader().Extract(context.Reader.GetBytes(valueOffset, byteCount)));
```
drewnoakes, line 43: Could the byte array be pooled here?

drewnoakes (commented):
Looks good. Left a bunch of comments that are mostly for future stuff. If you want to tackle any of them here, but all means do, otherwise I'll file follow-up items. Let me know when you think this is ready and I'll take another look (the description is WIP so I assume you're still playing with it).

MetadataExtractor/Formats/Icc/IccReader.cs
iamcarbon: The _position needs to be mutated when using the sequential access methods, so this needs to remain ref.

MetadataExtractor/Formats/Photoshop/PhotoshopTiffHandler.cs
iamcarbon: Should be! Will update.

MetadataExtractor/Formats/Bmp/BmpReader.cs
iamcarbon: Will tackle in this PR. Will look something like:

```
byte[]? rentedBuffer = null;

var iccBytes = profileSize <= 256
    ? stackalloc byte[profileSize]
    : (rentedBuffer = ArrayPool.Shared.Rent(profileSize)).Slice(0, profileSize);

...

if (rentedBuffer != null)
{
    ArrayPool.Shared.Return(rentedBuffer);
}

```

MetadataExtractor/Formats/Icc/IccReader.cs
drewnoakes: Ah yes, makes sense 👍

MetadataExtractor/Formats/Bmp/BmpReader.cs
drewnoakes: I'm wondering if we can make the call site look like this:

```c#
using rental = BufferUtil.Rent(minimumSize: 1234);

var reader = new BufferReader(rental.Buffer);
```

I'm not familiar enough with `ref struct`/`scoped` here to know if this is possible but I think it is. I think we can also return a `struct` disposable and the compiler won't box it to invoke `Dispose` through the interface virtually.

If we had a pattern like that, we could look at making the stack allocation smarter. I read something somewhere about being able to test for stack space, which might mean the threshold could be dynamic. Even a threshold of 256 could blow the stack. However I don't know if such a capability exists, or if I just read a proposal for that.

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -123,7 +108,7 @@ static object HandleInt(ref BufferReader reader, byte marker)
 
         static Dictionary<byte, byte> HandleDict(ref BufferReader reader, byte count)
         {
-            var keyRefs = ArrayPool<byte>.Shared.Rent(count);
+            Span<byte> keyRefs = stackalloc byte[count]; // count is < 256
 
             for (int j = 0; j < count; j++)
             {
```
drewnoakes, line 114: This bit can be changed from:

```c#
for (int j = 0; j < count; j++)
{
    keyRefs[j] = reader.GetByte();
}
```

To

```c#
reader.GetBytes(keyRefs);
```

iamcarbon: We're relying on an implicit upcast from byte to int here -- so we can't read directly. However, let me refactor to avoid the extra calls / and bounds checks.

MetadataExtractor/Formats/Apple/BplistReader.cs
iamcarbon: We're relying on an implicit upcast from byte to int here -- so we can't read directly. However, let me refactor to avoid the extra calls / bounds checks.

MetadataExtractor/Formats/Bmp/BmpReader.cs
```diff
@@ -347,8 +347,13 @@ private static void ReadBitmapHeader(SequentialReader reader, BmpHeaderDirectory
                         }
                         else
                         {
-                            var iccReader = new ByteArrayReader(reader.GetBytes(profileSize));
-                            var iccDirectory = new IccReader().Extract(iccReader);
+                            using var iccBuffer = profileSize <= 256
+                                ? new BufferScope(stackalloc byte[profileSize])
+                                : new BufferScope(profileSize);
```
drewnoakes, line 352: Running this over the test suite shows a `StackOverflowException` here. Were you able to get the regression suite working locally? Let me know if you need a hand with that.

The issue is that `profileSize` is negative.

Perhaps something like:

```suggestion
                            using var iccBuffer = (uint)profileSize <= 256
                                ? new BufferScope(stackalloc byte[profileSize])
                                : new BufferScope(profileSize);
```

That's a clever and performant way to validate the value is also greater than zero, but it's a less common pattern and doesn't read very naturally.

Wish there was a better way to wrap this pattern up, but I can't see it, at least not for all target frameworks we have. For .NET 8 I think we could use inline arrays with something like this:

```c#
internal readonly ref struct BufferScope
{
    private const int StackBufferSize = 256;

    private readonly Buffer _buffer;
    private readonly byte[]? _array;
    private readonly Span<byte> _span;

    public BufferScope(int size)
    {
        if (size < 0)
        {
            throw new ArgumentOutOfRangeException(nameof(size), size, "Cannot be negative.");
        }
        else if (size < StackBufferSize)
        {
            _span = MemoryMarshal.CreateSpan(ref Unsafe.As<Buffer, byte>(ref _buffer), size);
        }
        else
        {
            _array = ArrayPool<byte>.Shared.Rent(size);
            _span = _array.AsSpan(0, size);
        }
    }

    public readonly Span<byte> Span => _span;

    public readonly void Dispose()
    {
        if (_array is null) return;

        ArrayPool<byte>.Shared.Return(_array);
    }

    [InlineArray(StackBufferSize)]
    private struct Buffer
    {
        [SuppressMessage("CodeQuality", "IDE0051:Remove unused private members")]
        private byte _element0;
    }
}
```

MetadataExtractor/IO/BufferScope.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+using System.Buffers;
+
+namespace MetadataExtractor.IO;
+
+internal ref struct BufferScope
```
drewnoakes, line 7: Could add implicit conversion to a `Span<byte>` with:

```c#
    public static implicit operator Span<byte>(BufferScope bufferScope)
    {
        return bufferScope._span;
    }
```

That could clean up the usages a little. I'd also consider naming this `ScopedBuffer` as I think it's more of a buffer than a scope.

drewnoakes:
I'm seeing a few regressions in the image suite. One I called out above (stack overflow on negative size), and another seems to be in ICC data, as quite a few files now give:

> DOTNET [ERROR: ICC Profile] Exception reading ICC profile: Attempt to read from beyond end of underlying data source (requested index: 0, requested count: 4, max index: -1)

Let me know if you need a hand getting the regression suite set up on your machine. If you're not touching the Java version, you only need to set up the .NET runner (and can comment out the call to run the Java lib too).

iamcarbon:
Getting that setup now.

Screenshot 2024-02-09 at 8 08 24 PM

drewnoakes:
@iamcarbon let me know if you'd like me to pick this up. No real rush, but I'd like to get 2.9 out in the next couple of weeks.

markegorman:
I don’t know what this is or why I’m getting it

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

iamcarbon:
@drewnoakes OK, I got the regression suite working. Couple regressions I still need to figure out.

I'm going to break this PR apart few simpler PRs, and may back out of on the ICC span changes since this changes some error messages. I'll get that done this week. 

Screenshot 2024-02-18 at 7 11 33 PM

Screenshot 2024-02-18 at 7 11 23 PM

drewnoakes:
Sounds good. I suggest also running the suite without your changes, just to baseline. It's possible that different environmental factors create a diff. I try to minimise that, but it's hard.

iamcarbon:
Closing this in favor of smaller incremental PRs.

PR #415, opened 2024-02-22
Introduce ScopedBuffer

iamcarbon:
This PR extracts the ScopedBuffer work from #406

Local regression testing shows no diffs.

@drewnoakes Ready for review.

### DIFF
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -121,7 +121,7 @@ public sealed class BplistReader
 
         static Dictionary<byte, byte> HandleDict(ref BufferReader reader, byte count)
         {
-            Span<byte> keyRefs = stackalloc byte[count]; // <= 255
+            Span<byte> keyRefs = stackalloc byte[count]; // count is a byte (0-255)
 
             reader.GetBytes(keyRefs);
 
--- a/MetadataExtractor/Formats/Apple/BplistReader.cs
+++ b/MetadataExtractor/Formats/Apple/BplistReader.cs
@@ -1,7 +1,5 @@
 ﻿// Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
-using System.Buffers;
-
 namespace MetadataExtractor.Formats.Apple;
 
 /// <summary>
@@ -123,12 +121,9 @@ public sealed class BplistReader
 
         static Dictionary<byte, byte> HandleDict(ref BufferReader reader, byte count)
         {
-            var keyRefs = ArrayPool<byte>.Shared.Rent(count);
+            Span<byte> keyRefs = stackalloc byte[count]; // <= 255
 
-            for (int j = 0; j < count; j++)
-            {
-                keyRefs[j] = reader.GetByte();
-            }
+            reader.GetBytes(keyRefs);
 
             Dictionary<byte, byte> map = [];
 
@@ -136,8 +131,6 @@ public sealed class BplistReader
             {
                 map.Add(keyRefs[j], reader.GetByte());
             }
-
-            ArrayPool<byte>.Shared.Return(keyRefs);
 
             return map;
         }
--- a/MetadataExtractor/IO/BufferReader.Indexed.cs
+++ b/MetadataExtractor/IO/BufferReader.Indexed.cs
@@ -1,6 +1,5 @@
 // Copyright (c) Drew Noakes and contributors. All Rights Reserved. Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
 
-using System.Buffers;
 using System.Buffers.Binary;
 
 namespace MetadataExtractor.IO;
--- a/MetadataExtractor/IO/BufferReader.Indexed.cs
+++ b/MetadataExtractor/IO/BufferReader.Indexed.cs
@@ -193,17 +193,11 @@ internal ref partial struct BufferReader
         }
         else
         {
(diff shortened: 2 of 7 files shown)


MetadataExtractor/Util/DateUtil.cs
```diff
@@ -18,7 +18,7 @@ public static bool IsValidTime(int hours, int minutes, int seconds)
                minutes is >= 0 and < 60
                && seconds is >= 0 and < 60;
 
-#if NET8_0 || NETSTANDARD2_1
+#if NET8_0_OR_GREATER || NETSTANDARD2_1
```
drewnoakes, line 21: Thanks for fixing this. I searched and couldn't find any other instances.

MetadataExtractor/Formats/Apple/BplistReader.cs
```diff
@@ -123,12 +121,9 @@ static object HandleInt(ref BufferReader reader, byte marker)
 
         static Dictionary<byte, byte> HandleDict(ref BufferReader reader, byte count)
         {
-            var keyRefs = ArrayPool<byte>.Shared.Rent(count);
+            Span<byte> keyRefs = stackalloc byte[count]; // <= 255
```
drewnoakes, line 124: Could we convert the comment into a `Debug.Assert(count <= 255)` instead? It provides a little more safety and is clearer as far as documenting intent.

MetadataExtractor/IO/BufferReader.Indexed.cs
```diff
@@ -178,6 +177,9 @@ namespace MetadataExtractor.IO;
 
     public readonly string GetString(int index, int bytesRequested, Encoding encoding)
     {
+        if (bytesRequested < 0)
+            throw new ArgumentOutOfRangeException(nameof(bytesRequested), "Must be 0 or greater");
```
drewnoakes, line 181: super nit: exception messages should end with a full stop in .NET.

MetadataExtractor/IO/BufferReader.Indexed.cs
```diff
@@ -193,17 +195,11 @@ namespace MetadataExtractor.IO;
         }
         else
```
drewnoakes, line 196: Do we need the else block any more? Looks like the size check could be moved into a ternary conditional and the two blocks collapsed into one, as you've done elsewhere in the PR.

MetadataExtractor/IO/BufferReader.Sequential.cs
```diff
@@ -114,17 +114,19 @@ public ulong GetUInt64()
 
     public string GetString(int bytesRequested, Encoding encoding)
     {
-        // This check is important on .NET Framework
```
drewnoakes, line 117: Could we keep this comment on the zero check please? It still exists in other methods.

MetadataExtractor/IO/BufferReader.Sequential.cs
```diff
@@ -114,17 +114,19 @@ public ulong GetUInt64()
...
+        if (bytesRequested < 0)
+            throw new ArgumentOutOfRangeException(nameof(bytesRequested), "Must be 0 or greater");
+
         if (bytesRequested is 0)
             return "";
 
-        Span<byte> bytes = bytesRequested <= 256
-            ? stackalloc byte[bytesRequested]
-            : new byte[bytesRequested];
+        using var buffer = bytesRequested <= 256
```
drewnoakes, line 123: What do you think about moving the 256 to a constant, like `ScopedBuffer.MaxStackBufferSize`?

MetadataExtractor/IO/ScopedBuffer.cs
```diff
@@ -0,0 +1,43 @@
...
+        return bufferScope._span;
+    }
+
+    public void Dispose()
+    {
+        if (_rentedBuffer is null) return;
+
+        ArrayPool<byte>.Shared.Return(_rentedBuffer);
+
+        _rentedBuffer = null;
```
drewnoakes, line 41: I don't think we need to zero this field out. This type will be disposed when it falls out of scope, at which point it shouldn't be rooting the value anymore.

drewnoakes (approved):
Looks great. I like this pattern a lot. Much nicer than repeating ourselves all over the place.

Left a few comments for you to consider. Have approved, so merge it whenever you're happy with it, or let me know if you want me to take another look later on.

MetadataExtractor/IO/ScopedBuffer.cs
iamcarbon: originally added this to prevent a double disposal, but the uses are trivial and safe. removed.

MetadataExtractor/IO/BufferReader.Indexed.cs
iamcarbon: unified the logic across all the readers. much tidier.

MetadataExtractor/IO/BufferReader.Indexed.cs
iamcarbon: noted and fixed!

MetadataExtractor/Formats/Apple/BplistReader.cs
iamcarbon: I updated the comment to be clearer (count is an unsigned byte, with a safe range) -- no need for an assert. The comment saves a second when auditing the stackalloc call sites.

MetadataExtractor/Util/DateUtil.cs
iamcarbon: Yep, also searched -- and only one I found!

iamcarbon:
Addressed all feedback, and double checked against regressions. Merged!

drewnoakes:
Looks great, thanks.

FYI I will be less online over the next week but will do my best to review any PRs.

PR #422, opened 2024-05-05
Fixed PNG compressed iTXt chunk

### LINKED ISSUE #421: Bug in PNG compressed iTXt chunk
There seems to be a bug in the decoding of compressed iTXt chunks in PNG files. Instead of trying to decompress the actual compressed data stored after the keyword and flags, the data from the start of the whole chunk is sent instead which causes a failure:
https://github.com/drewnoakes/metadata-extractor-dotnet/blob/ed79cc04b915173e1c27318ad90194a3f9caf515/MetadataExtractor/Formats/Png/PngMetadataReader.cs#L267

I have fixed the issue in a branch here and will create a PR:
https://github.com/reinfallt/metadata-extractor-dotnet/tree/png_iTXt_compressed

Here is an example png file: 
red_itxt

reinfallt:
Fixes #421

### DIFF
--- a/MetadataExtractor/Formats/Png/PngMetadataReader.cs
+++ b/MetadataExtractor/Formats/Png/PngMetadataReader.cs
@@ -264,7 +264,9 @@ namespace MetadataExtractor.Formats.Png
                 {
                     if (compressionMethod == 0)
                     {
-                        if (!TryDeflate(bytes, bytesLeft, out textBytes, out string? errorMessage))
+                        reader.Skip(2); // Skip over the zlib header bytes (78 9C)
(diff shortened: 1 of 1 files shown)


drewnoakes:
Looks great thanks. Please bear with me as I'll need to pull this down and run over the regression suite.

drewnoakes:
@reinfallt apologies for the long delay on this, but do you possibly have an image that hits this code path? We don't have one in the regression suite. Would be great if you could share one.

drewnoakes:
Thank you very much @reinfallt. I'm trying to get a release out soon.

Issue #344, opened 2023-09-02
Reading iPhone MOV throws File format could not be determined


rnedumpurath:
Starting around iOS15, I've been getting the "File format could not be determined " exception when trying to read the metadata of a video file of a Live Photo. A Live Photo on an iPhone creates an HEIC image file as well as a video file. The video file from that package is what is throwing the exception. I'm attaching a sample video file that should recreate this exception. Tested with .NET 7 running MetadataExtractor 2.8.1.

https://github.com/drewnoakes/metadata-extractor-dotnet/assets/60277804/909ef10e-fcb9-4e62-8c5a-d5b62962450e

HEIC-to-JPEG-Dev:
Yes, I get this as well. Apple don’t like Windows :(

I’d use this library if this was fixed, otherwise it’s of no use to me as Apple files are prolific.

I can provide video files if that helps with the development of this capability

HEIC-to-JPEG-Dev:
I would add that “throwing an exception” is very bad for performance. If I’m getting the data for a thousand files and 20% are Apple videos, this package hits a wall performance wise. It should return a code instead. Exceptions are for unexpected errors if all goes wrong, not for known, expected things you’re checking for.

HEIC-to-JPEG-Dev:
I think Apple are using hevc encoding now

naughtyGitCat:
same execption on live image mov too
```
MetadataExtractor.ImageProcessingException: File format could not be determined
   at MetadataExtractor.ImageMetadataReader.ReadMetadata(Stream stream)
   at MetadataExtractor.ImageMetadataReader.ReadMetadata(String filePath)
```


https://github.com/user-attachments/assets/8d126a7f-f4fa-4ca2-88ba-06776e8c21c0

naughtyGitCat:
@rn-89 @HEIC-to-JPEG-Dev   happens with same error, the header info changes, fix it with an `if .... else ....` to  catch this circumstance

@drewnoakes please take a notice, can you merge my pr?

HEIC-to-JPEG-Dev:
I'm guessing the repro is dead now ?

drewnoakes:
Thanks @naughtyGitCat for fixing this. Your PR has merged, and a new release will go out shortly.

PR #426, opened 2024-09-09
add support for apple live-photo mov file refer #344

naughtyGitCat:
Live Photos is actually two files. Original JPEG/HEIF Image and Full HD Video(QuickTime). In the ios ecology, live-photo connect image with mov with same name, and meta key `Content Identifier`

If we export the live photo mov file with  apple `afcservice` and `afcclient` via native protocol, the mov file lack `ftyp` in the file header

```
            // for live photo which is export from  native protocols with afcclient/afcservice
            // example 1
            // 00000000: 0000 0008 7769 6465 0033 a50e 6d64 6174  ....wide.3..mdat
            // 00000010: 0000 001f 4e01 051a 4756 4adc 5c4c 433f  ....N...GVJ.\LC?
            // 00000020: 94ef c511 3cd1 43a8 03ee 1111 ee02 00c3  ....<.C.........
            // example 2
            // 00000000: 0000 0008 7769 6465 004f 854d 6d64 6174  ....wide.O.Mmdat
            // 00000010: 0000 002a 4e01 0523 4756 4adc 5c4c 433f  ...*N..#GVJ.\LC?
            // 00000020: 94ef c511 3cd1 43a8 0000 0300 0003 0000  ....<.C.........
```

the pattern is `....wide....mdat`

ref:
- https://developer.apple.com/library/archive/releasenotes/General/WhatsNewIniOS/Articles/iOS9_1.html#//apple_ref/doc/uid/TP40016572-DontLinkElementID_2
- https://developer.apple.com/documentation/quicktime-file-format/quicktime_movie_files


drewnoakes:
Please remove the test data file and unit tests. We use a regression test suite to maintain quality.

naughtyGitCat:
> Please remove the test data file and unit tests. We use a regression test suite to maintain quality.

test data file and unit tests removed, and revert to other non-related changes

drewnoakes (approved):
I merged in main, moved the test image to the regression suite, and ran this change across that suite. It only triggered a change on the file you provided, and the output looks great.

Thanks again for fixing this! Will get a release out soon.

PR #427, opened 2024-10-07
SequentialStreamReader.cs: Cast numbers to ulong before bound check

### LINKED ISSUE #428: Integer overflow in SequentialStreamReader causes wrong length check
I have a corrupt image which goes into an infinite loop when running Metadata Extractor on it.
Some of the bad code is on my end (bad implementation of `Stream`) but one of the fail points is also in Metadata Extractor.

In the `SequentialStreamReader` class, The `Skip` method does a bounds check on `position + n > length`:  

image  

In my case, `n` (num bytes to skip) was very large but still positive (top-most bit is off) `0x7fffffffffffffef = 9223372036854775791 `
and the addition of `position + n` lit the top-most bit, so the result is a negative long `0x8000000000014b86 = -9223372036854690938`

Since both sides of the comparison are `long` and `-9223372036854690938 < 131072` then the bounds violation is not detected.

I made a PR to fix this issue: simply casting both sides of the comparison to `ulong` should solve it:
https://github.com/drewnoakes/metadata-extractor-dotnet/pull/427

I hope this helps.

KirkLandTrip:
Fixes #428

Skip Method: Avoid an integer overflow bug by explicitly casting to ulong before comparing the expression.

### DIFF
--- a/MetadataExtractor/IO/SequentialStreamReader.cs
+++ b/MetadataExtractor/IO/SequentialStreamReader.cs
@@ -85,7 +85,7 @@ namespace MetadataExtractor.IO
             if (n < 0)
                 throw new ArgumentException("n must be zero or greater.");
 
-            if (_stream.Position + n > _stream.Length)
+            if ((ulong)(_stream.Position + n) > (ulong)(_stream.Length))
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
Looks great, thanks very much!

PR #433, opened 2025-02-09
QuickTime: Fix parsing of the GPS tag (xyz)

### LINKED ISSUE #432: QuickTime: Regression in parsing GPS Location
Hey,
I used to work with Metadata Extractor in version 2.7.x and got QuickTime locations values.  
I switched to the latest code from `main` and the locations aren't parsed anymore.

I traced the error to this line in `QuickTimeMetadataReader`, in the `UserDataHandler` function:  
Image

It's trying to read the "type" as a string and it comes out as `"©xyz"`. The code is expecting `?xyz`.  
The binary value is 0xa978797a and the `0xa9` byte is the culprit. Since it's defined "outside of the ASCII" range, it gets interpreted differently in some cases.  
I'm not sure if my locale (en-GB) or switching from .NET framework to .NET 7 cause this, but at least some users beside me are going to experience it.

KirkLandTrip:
Fixes #432

### DIFF
--- a/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
+++ b/MetadataExtractor/Formats/QuickTime/QuickTimeMetadataReader.cs
@@ -122,9 +122,9 @@ namespace MetadataExtractor.Formats.QuickTime
 
             void UserDataHandler(AtomCallbackArgs a)
             {
-                switch (a.TypeString)
+                switch (a.Type)
                 {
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
Thanks very much for this. Looks great. Pushed a fix for an issue after merging in the latest main branch.

drewnoakes:
Will be getting a new release out soon.

Issue #453, opened 2025-09-05
UTC Offset exception reading a thermal image


dave-murray:
I have a .NET 9 for Android app that reads the FLIR EXIF data from thermal images using MetadataExtractor. One of my sample images throws an exception from MetadataExtractor but works in EXIF Tool and the FLIR SDK.

I do notice a warning in the EXIF Tools output - Bad format (65535) for IFD1 entry 6 - but I have no idea what that might mean.

Edit: I should have mentioned, this is using MetadataExtractor v2.9.0-rc3.

```
{System.ArgumentException: The UTC Offset for Utc DateTime instances must be 0. (Parameter 'offset')
   at System.DateTimeOffset..ctor(DateTime dateTime, TimeSpan offset)
   at MetadataExtractor.Formats.Flir.FlirReader.Extract(IndexedReader reader)+MoveNext()
   at System.Collections.Generic.List`1[[MetadataExtractor.Directory, MetadataExtractor, Version=2.9.0.0, Culture=neutral, PublicKeyToken=b66b5ccaf776c301]].AddRange(IEnumerable`1 collection)
   at MetadataExtractor.Formats.Jpeg.JpegMetadataReader.ProcessJpegSegments(IEnumerable`1 readers, ICollection`1 segments)
   at MetadataExtractor.Formats.Jpeg.JpegMetadataReader.Process(Stream stream, ICollection`1 readers)
   at MetadataExtractor.Formats.Jpeg.JpegMetadataReader.ReadMetadata(Stream stream, ICollection`1 readers)
   at MetadataExtractor.ImageMetadataReader.ReadMetadata(Stream stream, String fileName)
   at MetadataExtractor.ImageMetadataReader.ReadMetadata(String filePath)
   at ClarityWORK.Droid.Services.FlirImageProcessingServiceMobile.ReadExifData(String filePath)
```

This exception is thrown by the line:
```c#
        IReadOnlyList<MetadataExtractor.Directory> directories = ImageMetadataReader.ReadMetadata(filePath);
```

The `filePath` is correct and other FLIR images load and read fine.

Problem image: (FLIR0141.jpg)
![Image](https://github.com/user-attachments/assets/1a0238b8-624a-48c9-a887-d912915df078)

Working image: (FLIR2541.jpg)
![Image](https://github.com/user-attachments/assets/3a739cfb-ebcd-4ff6-a843-d0d9a5def415)

I have more working images available if required.

[...]

PR #458, opened 2025-09-24
Use platform method for Unix time conversion

dave-murray:
This fixes issue #453 UTC Offset exception reading a thermal image.

All the unit tests pass, I haven't run regression testing because I don't have the ability to build the Java library. I've also tested it on 25 sample FLIR images provided by our engineers, most of them taken not in UTC.

`DateTimeOffset.FromUnixTimeSeconds` has been availbale since .Net Framework v4.6 and is in .NET Standard 2.0 and 2.1.

### DIFF
--- a/MetadataExtractor/Formats/Flir/FlirReader.cs
+++ b/MetadataExtractor/Formats/Flir/FlirReader.cs
@@ -169,7 +169,9 @@ namespace MetadataExtractor.Formats.Flir
                     var tm = dateTimeReader.GetUInt32();
                     var ss = dateTimeReader.GetUInt32() & 0xffff;
                     var tz = dateTimeReader.GetInt16();
-                    directory.Set(TagDateTimeOriginal, DateTimeOffset.FromUnixTimeSeconds(tm - (tz * 60)).AddSeconds(ss / 1000d));
+                    directory.Set(TagDateTimeOriginal,
+                        new DateTimeOffset(DateTimeOffset.FromUnixTimeSeconds(tm - (tz * 60)).AddSeconds(ss / 1000d).Ticks,
+                        TimeSpan.FromMinutes(-tz)));
 
                     directory.Set(TagFocusStepCount, reader2.GetUInt16(TagFocusStepCount));
                     directory.Set(TagFocusDistance, reader2.GetFloat32(TagFocusDistance));


drewnoakes:
Thanks for this! Does this mean we can remove `DateUtil.FromUnixTime`?

FWIW you don't have to build the Java lib to run the regression suite. Just clone the (admittedly large) https://github.com/drewnoakes/metadata-extractor-images and run https://github.com/drewnoakes/metadata-extractor-images/blob/main/src/dotnet/MetadataExtractor.MediaLibraryProcessor/Program.cs. You can comment out the `JavaRunner` and `DiffRunner` and see whether any of the `metadata/dotnet` files change. Note that some changes can be due to different versions of .NET or other environmental factors. I've tried to reduce that as much as possible, but a safe way to work around that is to run the tool without your changes, stage any changes to metadata files in git, then run again with your changes and see if further changes were added.

dave-murray:
> Thanks for this! Does this mean we can remove `DateUtil.FromUnixTime`?

Yes it does. I had intended to do so but forgot, sorry. Oversight corrected.

I'll have a look at the regression tests now.

dave-murray:
Regression tests showed I was losing the time zone info, this hadn't been apparent with my test images because I'm in the same time zone. I've updated the code to add the time zone.

Running the regression tests again the time zone has changed sign and I think this is correct.

FILE: FLIR iPhone device.jpg
Original: [FLIR Camera Info - 0x0384] Date Time Original = 16/05/2021 12:27:36 +07:00
New: [FLIR Camera Info - 0x0384] Date Time Original = 16/05/2021 12:27:36 -07:00
Exif Tool: DateTimeOriginal = 2021:05:16 12:27:36.107-07:00

FILE: FLIR Vue Pro 640.jpg
Original: [FLIR Camera Info - 0x0384] Date Time Original = 27/06/2016 15:29:42 +04:00
New: [FLIR Camera Info - 0x0384] Date Time Original = 27/06/2016 15:29:42 -04:00
Exif Tool: DateTimeOriginal = 2016:06:27 15:29:42.041-04:00

The new code also shows all my test images are UTC or UTC+1 which is correct.

The results for these two files are the only ones that have changed.

drewnoakes (approved):
Thanks very much for the update @dave-murray. This looks great.

There's no FLIR support on the Java side, so we don't need to worry about regression runs there or parity.

